PR #1, opened 2017-06-05
Initial main

dgromov:


.travis.yml
```diff
@@ -10,5 +10,4 @@ dist: trusty
 script:
  - make lint
```
prateek: Follow the convention described [here](https://github.com/m3db/ci-scripts) for the makefile. 

use `make all` in the .travis.yml like other M3DB repos: https://github.com/m3db/m3db/blob/master/.travis.yml

handler/common.go
```diff
@@ -0,0 +1,55 @@
...
+
+// MethodNotSupportedResponse writes a method not supported response to the writer
+func MethodNotSupportedResponse(w http.ResponseWriter, r *http.Request, supported []string) {
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	response := NotSupportdResponse{
+		Error:            "Method not supported",
+		URI:              r.RequestURI,
+		MethodName:       r.Method,
+		SupportedMethods: supported}
+	body, _ := json.Marshal(response)
```
prateek: nit: handle the error - propagate up/panic

handler/common_test.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+
+	// Our handlers satisfy http.Handler, so we can call their ServeHTTP method
+	// directly and pass in our Request and ResponseRecorder.
+	handler.ServeHTTP(rr, &req)
+
+	rawResult := make([]byte, rr.Body.Len())
+	_, err := rr.Body.Read(rawResult)
+	if err != nil {
+		t.Errorf("Encountered error parsing reponse")
```
prateek: shorter:

```
require.NoError(t, err, "Encountered error parsing response)
```

handler/health/health.go
```diff
@@ -0,0 +1,89 @@
...
+	host := gatherHostInfo()
+	h := health{Host: host, Timestamp: start}
+	status := healthCheck()
+	h.ResponseTime = time.Since(start)
+
+	h.Status = status
+	if status != ok {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+	body, _ := json.Marshal(h)
```
prateek: nit: handle error

handler/health/health.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Returns health check status msg plus additional message
+func healthCheck() string {
+	return ok
+}
+
+func gatherHostInfo() string {
+	host, err := os.Hostname()
+	if err != nil {
+		host = "unknown"
```
prateek: nit: why not propagate the error up to the `m3ctlHealthCheck`? 

Use a const instead of a magic string here

handler/health/health.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/m3db/m3x/instrument"
+)
+
+const (
+	ok   = "OK"
+	fail = "FAIL"
+)
+
+var (
+	healthURL = "/health"
```
prateek: any reason to make this a var instead of a const?

handler/health/health.go
```diff
@@ -0,0 +1,89 @@
...
+package health
+
+import (
+	"encoding/json"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3x/instrument"
```
prateek, line 31: nit: blank line between repo local and third-party imports

handler/rules/rules.go
```diff
@@ -0,0 +1,81 @@
...
+package rules
+
+import (
+	"net/http"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3x/instrument"
+)
+
+var (
```
prateek: nit: can these be `const`?

handler/rules/rules.go
```diff
@@ -0,0 +1,81 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package rules
+
+import (
+	"net/http"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3x/instrument"
```
prateek, line 27: nit: empty line between these imports

handler/rules/global.go
```diff
@@ -0,0 +1,39 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package rules
+
+import "net/http"
```
prateek: nit: import block

server/options.go
```diff
@@ -0,0 +1,98 @@
...
+	opts.instrumentOpts = value
+	return &opts
+}
+
+func (o *options) InstrumentOptions() instrument.Options {
+	return o.instrumentOpts
+}
+
+// NewOptions creates a new set of server options
+func NewOptions() Options {
```
prateek: nit: reorder so the `ctor` is not in the middle of getters/setters

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+	etcdclient "github.com/m3db/m3cluster/client/etcd"
+	"github.com/m3db/m3ctl/server"
+	"github.com/m3db/m3ctl/services/m3ctl/serve"
+	"github.com/m3db/m3x/config"
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
+)
+
+type serverConfig struct {
```
prateek, line 38: Feels weird to have the struct be internal but the NewHTTPServerOptions method be public. Either export both, or neither. 

If you're going to export the config struct, move to another package so other mains can import it

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,148 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	etcdclient "github.com/m3db/m3cluster/client/etcd"
+	"github.com/m3db/m3ctl/server"
+	"github.com/m3db/m3ctl/services/m3ctl/serve"
```
prateek: nit: import ordering and spacing

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if c.WriteTimeout != 0 {
+		opts = opts.SetWriteTimeout(c.WriteTimeout)
+	}
+
+	return opts
+}
+
+type kvClientConfiguration struct {
+	Etcd *etcdclient.Configuration `yaml:"etcd"`
```
prateek: any reason to define this instead of not using etcdclient.Configuration directly?

server/server.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package server
```
prateek, line 21: Can you re-use the newly added `m3x/server` package?

server/handlers.go
```diff
@@ -0,0 +1,44 @@
...
+
+package server
+
+import (
+	"net/http"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3ctl/handler/health"
+	"github.com/m3db/m3ctl/handler/rules"
+	"github.com/m3db/m3x/instrument"
```
prateek: nit: space above this line

handler/health/health.go
dgromov: I took this out of go-common. Its not really an issue, if it is unknown its fine.

handler/health/health.go
dgromov: Just following the pattern we have everywhere else

services/m3ctl/main/main.go
dgromov: to not pollute the top level config namespace

server/server.go
dgromov: Will look into this.

glide.yaml
```diff
@@ -0,0 +1,11 @@
+package: github.com/m3db/m3ctl
+
+import:
```
jeromefroe, line 3: Maybe copy the packages (and their corresponding versions) that are in `glide.lock` but aren't in `glide.yaml` into `glide.yaml` so they don't change when someone runs `glide update`?

glide.yaml
```diff
@@ -0,0 +1,11 @@
...
+- package: github.com/m3db/m3cluster
+  version: d64606edb68599677fd335ceed86d990d13d2991
+- package: github.com/m3db/m3metrics
+  version: 18bc01a7057d55b78e528236e0e9db297bcd4fa3
+- package: github.com/m3db/m3x
+  version: bd6be18ffb184691512eb442561b16a59c15d797
+- package: github.com/stretchr/testify
```
jeromefroe: nit: could mark this under `TestImports`. For example:

```
testImport:
- package: github.com/stretchr/testify
  version: 6fe211e493929a8aac0469b93f28b1d0688a9a3a
```

handler/common.go
```diff
@@ -0,0 +1,60 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+const (
+	mimeTypeApplicationJSON = "application/json"
+)
+
+// NotSupportdResponse is data for a method not supported response
```
jeromefroe: s/NotSupportdResponse/NotSupportedResponse

handler/common.go
```diff
@@ -0,0 +1,60 @@
...
+type NotSupportdResponse struct {
+	Error            string
+	MethodName       string
+	URI              string
+	SupportedMethods []string
+}
+
+// MethodNotSupportedResponse writes a method not supported response to the writer
```
jeromefroe: nit: perhaps add periods to the end of these comments. It's the comment convention recommended in the Go wiki and we've been trying to migrate code to use it, though we're definitely far from that mark :) Can start clean here at least

handler/controller.go
```diff
@@ -0,0 +1,29 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE
+
+package handler
+
+import "net/http"
+
+// Controller defines functions around a controllable entity
+type Controller interface {
+	// Registers the handlers for this controll with the given ServeMux
```
jeromefroe: s/controll/contoller

nit: comments should ideally start with the name of the thing that is being described. So perhaps "RegisterHandlers registers the handlers for this controller with the given ServeMux"?

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+var ()
+
+type health struct {
+	Host         string        `json:"host"`
+	Timestamp    time.Time     `json:"timestamp"`
+	ResponseTime time.Duration `json:"response_time"`
+	Status       string        `json:"status"`
+}
+
+// Controller sets up the handlers for health endpoint
```
jeromefroe: s/Controller/controller

handler/common.go
```diff
@@ -0,0 +1,60 @@
...
+func MethodNotSupportedResponse(w http.ResponseWriter, r *http.Request, supported []string) error {
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	response := NotSupportdResponse{
+		Error:            "Method not supported",
+		URI:              r.RequestURI,
+		MethodName:       r.Method,
+		SupportedMethods: supported}
+	body, err := json.Marshal(response)
+	if err != nil {
```
jeromefroe: perhaps just return the error itself or include it in the error message?

handler/common_test.go
```diff
@@ -0,0 +1,62 @@
...
+		MethodNotSupportedResponse(w, r, goodMethods)
+	})
+
+	// Our handlers satisfy http.Handler, so we can call their ServeHTTP method
+	// directly and pass in our Request and ResponseRecorder.
+	handler.ServeHTTP(rr, &req)
+
+	rawResult := make([]byte, rr.Body.Len())
+	_, err := rr.Body.Read(rawResult)
+	require.NoError(t, err, "Encountered error parsing reponse")
```
jeromefroe: s/reponse/response

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+type controller struct {
+	instruments instrument.Options
+}
+
+// NewController creates a new rules controller
+func NewController(instrument instrument.Options) handler.Controller {
+	return &controller{instruments: instrument}
+}
+
+// Returns health check status msg plus additional message
```
jeromefroe: same as the note above, perhaps start this comment with "healthCheck ..."

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+
+func gatherHostInfo() string {
+	host, err := os.Hostname()
+	if err != nil {
+		host = unknownName
+	}
+	return host
+}
+
+// M3CtlHealthCheck is a health check for the m3ctl server
```
jeromefroe: s/M3CtlHealthCheck/m3CtlHealthCheck

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+	body, _ := json.Marshal(h)
+
+	w.Write(body)
+}
+
+// RegisterHandlers registers health handlers
+func (c *controller) RegisterHandlers(mux *http.ServeMux) error {
+	log := c.instruments.Logger()
```
jeromefroe: nit: add a `log` field to `controller` so that you can just call `c.log.Infof(...)`

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package rules
+
+import (
+	"net/http"
+
+	"github.com/m3db/m3x/instrument"
+
```
jeromefroe: nit: extra blank space

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/m3db/m3ctl/handler"
+)
+
+const (
+	getAllURL          = "/rules/namespaces"
+	createOrDeleteURL  = "/rules/namespace"
+	modifyNamespaceURL = "/rules/namespace/"
+)
+
+// Controller sets up the handlers for rules
```
jeromefroe: s/Controller/controller

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+type controller struct {
+	instruments instrument.Options
```
jeromefroe: nit: iOpts might be a more succinct field name than `instruments` and is commonly used in some of the other repos

server/handlers.go
```diff
@@ -0,0 +1,36 @@
...
+// THE SOFTWARE
+
+package server
+
+import (
+	"net/http"
+
+	"github.com/m3db/m3ctl/handler/health"
+	"github.com/m3db/m3ctl/handler/rules"
+
```
jeromefroe: nit: extra blank space

server/server.go
jeromefroe: Or perhaps the `Server` struct defined in the `http` package since this will be an HTTP server: https://golang.org/pkg/net/http/#Server

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/m3db/m3x/config"
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
+
+	"github.com/m3db/m3ctl/server"
+	"github.com/m3db/m3ctl/services/m3ctl/serve"
+)
+
+type serverConfig struct {
+	// HTTP server listening address.
```
jeromefroe: nit: perhaps starts the comment as "ListenAddress ..." for this field and the ones below

.travis.yml
```diff
@@ -8,7 +8,4 @@ env:
 sudo: required
 dist: trusty
 script:
- - make lint
- - make test-ci-unit
- - make test-ci-integration
- - make m3ctl
+ - make all
```
xichen2020: Add a newline here?

Makefile
```diff
@@ -42,6 +42,9 @@ services-linux-amd64:
 
 $(foreach SERVICE,$(SERVICES),$(eval $(SERVICE_RULES)))
 
+all: lint test-ci-unit services
+	@echo Made all successfully
```
xichen2020, line 46: It'll fail fairly loudly if any of the build targets fail so don't think you actually need the echo here

glide.yaml
xichen2020: +1

handler/health/health.go
dgromov: Its very unlikely that this will fail. Also, what do you suggest I do in this case?

handler/common.go
xichen2020: Comments should be sentences and end with period.

xichen2020: Also should it be called MethodNotSupportedResponse?

handler/common.go
```diff
@@ -0,0 +1,60 @@
...
+// NotSupportdResponse is data for a method not supported response
+type NotSupportdResponse struct {
+	Error            string
+	MethodName       string
+	URI              string
+	SupportedMethods []string
+}
+
+// MethodNotSupportedResponse writes a method not supported response to the writer
+func MethodNotSupportedResponse(w http.ResponseWriter, r *http.Request, supported []string) error {
```
xichen2020: Feels like this function should be broken into two parts, first part is constructing the response, and second part is to write out the response, which IMO is a clearer separation of concerns.

handler/common.go
```diff
@@ -0,0 +1,60 @@
...
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	response := NotSupportdResponse{
+		Error:            "Method not supported",
```
xichen2020: Make it a const?

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+package health
+
+import (
+	"encoding/json"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/m3db/m3ctl/handler"
+
```
xichen2020: nit: remove the blank line?

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+type health struct {
+	Host         string        `json:"host"`
+	Timestamp    time.Time     `json:"timestamp"`
+	ResponseTime time.Duration `json:"response_time"`
+	Status       string        `json:"status"`
+}
+
+// Controller sets up the handlers for health endpoint
+type controller struct {
+	instruments instrument.Options
```
xichen2020: The convention is to call this `instrumentOpts`

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+func NewController(instrument instrument.Options) handler.Controller {
+	return &controller{instruments: instrument}
+}
+
+// Returns health check status msg plus additional message
+func healthCheck() string {
+	return ok
+}
+
+func gatherHostInfo() string {
```
xichen2020: nit: call this `hostName`?

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		host = unknownName
+	}
+	return host
+}
+
+// M3CtlHealthCheck is a health check for the m3ctl server
+func m3ctlHealthCheck(w http.ResponseWriter, r *http.Request) {
+	handler.InitJSONResponse(w)
+	start := time.Now()
```
xichen2020: This probably should be at the beginning of the method. We also conventionally pass in a clock.Options object which provides the `now` function for easier testing purpose.

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+	body, _ := json.Marshal(h)
+
+	w.Write(body)
+}
+
+// RegisterHandlers registers health handlers
+func (c *controller) RegisterHandlers(mux *http.ServeMux) error {
+	log := c.instruments.Logger()
+	mux.HandleFunc(healthURL, m3ctlHealthCheck)
+	log.Infof("Registered health endpoints")
```
xichen2020: Probably better to log the url? `"Registered endpoint %s", healthURL`?

handler/health/health.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil
```
xichen2020: When will we return a non-nil error here?

handler/rules/global.go
```diff
@@ -0,0 +1,41 @@
...
+// THE SOFTWARE.
+
+package rules
+
+import (
+	"fmt"
+	"net/http"
+)
+
+func getNamespaces(w http.ResponseWriter, r *http.Request) {
```
xichen2020: The convention is to not put `get` in getter methods: https://golang.org/doc/effective_go.html#Getters. Perhaps just call it `namespaces`?

handler/rules/namespace.go
```diff
@@ -0,0 +1,36 @@
...
+// THE SOFTWARE.
+
+package rules
+
+import (
+	"fmt"
+	"net/http"
+)
+
+func getRulesInNamespace(w http.ResponseWriter, r *http.Request) {
```
xichen2020: `rulesInNamespace`?

handler/rules/global.go
```diff
@@ -0,0 +1,41 @@
...
+	w.WriteHeader(http.StatusNotImplemented)
+	fmt.Fprintf(w, "Placeholder for getNamespaces")
+}
+
+func createNamespace(w http.ResponseWriter, r *http.Request) {
```
xichen2020: It might be better to have a struct that wraps the response writer and the request and performs namespace/rule-specific logic for better encapsulation. It's a bit confusing to see global functions float around.

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"net/http"
+
+	"github.com/m3db/m3x/instrument"
+
+	"github.com/m3db/m3ctl/handler"
+)
+
+const (
+	getAllURL          = "/rules/namespaces"
```
xichen2020: Should this be `namespacesURL`?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	createOrDeleteURL  = "/rules/namespace"
```
xichen2020: `rulesURL`?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	modifyNamespaceURL = "/rules/namespace/"
```
xichen2020: hm how is this different from above?

handler/controller.go
xichen2020: I wonder if `Router` might be a better name?

handler/rules/rules.go
xichen2020: `instrumentOpts` is also commonly used

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	instruments instrument.Options
+}
+
+// NewController creates a new rules controller
+func NewController(instrument instrument.Options) handler.Controller {
+	return &controller{instruments: instrument}
+}
+
+func (c *controller) handleNotSupported(w http.ResponseWriter, r *http.Request, supported []string) {
+	err := handler.MethodNotSupportedResponse(w, r, []string{http.MethodGet})
```
xichen2020: Make the supported method list a global var?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+func (c *controller) RegisterHandlers(mux *http.ServeMux) error {
+	log := c.instruments.Logger()
+	mux.HandleFunc(getAllURL, c.handleGetAll)
+	mux.HandleFunc(createOrDeleteURL, c.handleCreateDelete)
+	mux.HandleFunc(modifyNamespaceURL, c.handleModify)
+	log.Infof("Registered rules endpoints")
+	return nil
+}
+
+func (c *controller) handleGetAll(w http.ResponseWriter, r *http.Request) {
```
xichen2020: `getNamespaces` might be a better name?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		c.instruments.Logger().Errorf("%v", err)
+	}
+}
+
+// RegisterHandlers registers rule handlers
+func (c *controller) RegisterHandlers(mux *http.ServeMux) error {
+	log := c.instruments.Logger()
+	mux.HandleFunc(getAllURL, c.handleGetAll)
+	mux.HandleFunc(createOrDeleteURL, c.handleCreateDelete)
```
xichen2020: We should probably register separate handlers for creation and deletion?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+	mux.HandleFunc(modifyNamespaceURL, c.handleModify)
```
xichen2020: We should probably register separate handlers for creation and deletion?

handler/rules/rules.go
```diff
@@ -0,0 +1,96 @@
...
+type controller struct {
+	instruments instrument.Options
+}
+
+// NewController creates a new rules controller
+func NewController(instrument instrument.Options) handler.Controller {
+	return &controller{instruments: instrument}
+}
+
+func (c *controller) handleNotSupported(w http.ResponseWriter, r *http.Request, supported []string) {
```
xichen2020: `notSupported` might be better?

server/handlers.go
xichen2020: Technically they fall under the same m3db organization so don't think we need the blank line here.

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,155 @@
...
+		close(closedCh)
+	}()
+
+	// Handle interrupts.
+	logger.Warnf("interrupt: %v", interrupt())
+
+	close(doneCh)
+
+	select {
+	case <-closedCh:
```
xichen2020: This unfortunately doesn't do what you want, as soon as `doneCh` is closed, the goroutine will proceed to close `closedCh` with no knowledge about whether the server is actually closed.

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,155 @@
...
+		logger.Info("server closed clean")
```
xichen2020: You probably also want a timeout here in case it takes too long to shut itself down.

Makefile
dgromov: We do this in other places, its nice positive reenforcement.

handler/common.go
dgromov: Writing the response is just w.Write(body). You want that in its own function?

handler/rules/global.go
dgromov: This is not an accessor method. This is a handler for getting all of the namespaces from etcd.

handler/rules/namespace.go
dgromov: same as before, not an accessor.

handler/rules/rules.go
dgromov: the slash

handler/rules/rules.go
dgromov: Its the same endpoint with different http verbs. Handlers are not verb specific

handler/health/health.go
jeromefroe: I agree with Prateek here, all errors should ideally be handled (whether thats performing some action, bubbling them up, panicking, etc.). In this case I think we can log the error and return a 500.

handler/rules/global.go
jeromefroe: The convention here is that handler methods should be of the form `..Handler`, see for example: https://golang.org/doc/articles/wiki/ and https://github.com/gorilla/mux#examples. So how about `namespacesHandler`?

handler/rules/namespace.go
jeromefroe: See above, maybe `rulesInNamespaceHandler`?

handler/rules/rules.go
jeromefroe: In that case, perhaps `modifyNamespaceURL = createOrDeleteURL + "/"` would be more appropriate?

handler/rules/rules.go
jeromefroe: +1

I think we should make the variable names a little more generic, for example `rulesURL` or `namespaceURL` carries less semantic information.

handler/rules/rules.go
jeromefroe: +1

glide.yaml
```diff
@@ -0,0 +1,66 @@
...
+- package: github.com/shirou/gopsutil
+  version: b62e301a8b9958eebb7299683eb57fab229a9501
+- package: github.com/shirou/w32
+  version: bb4de0191aa41b5507caa14b0650cdbddcd9280b
+- package: github.com/StackExchange/wmi
+  version: e542ed97d15e640bdc14b5c12162d59e8fc67324
+- package: github.com/uber-go/atomic
+  version: e682c1008ac17bf26d2e4b5ad6cdd08520ed0b22
+- package: github.com/uber-go/tally
+  version: 4b9d9de43ffcb0b7efe67dab2a92c2d58dbf16ab
```
prateek: `Tally` uses semantic versioning, change this to `<4.0.0`

glide.yaml
prateek: Only for direct dependencies, not transitive ones, yeah?

handler/health/health.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"encoding/json"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3x/instrument"
+
+	"fmt"
```
prateek: move this to be along side the other stdlib imports

handler/health/health.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	ok          = "OK"
+	fail        = "FAIL"
+	healthURL   = "/health"
+	unknownName = "unkown"
+)
+
+var ()
```
prateek: delete

handler/rules/rules.go
```diff
@@ -0,0 +1,57 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package rules
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/m3db/m3ctl/handler"
```
prateek: separate repo local import

server/handlers.go
```diff
@@ -0,0 +1,35 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE
+
+package server
+
+import (
+	"github.com/m3db/m3ctl/handler/health"
+	"github.com/m3db/m3ctl/handler/rules"
+	"github.com/m3db/m3x/instrument"
```
prateek, line 26: separate repo local v third-party imports

services/m3ctl/main/main.go
```diff
@@ -0,0 +1,147 @@
...
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/m3db/m3cluster/client/etcd"
+	"github.com/m3db/m3x/config"
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
+
+	"github.com/m3db/m3ctl/server"
```
prateek: move repo local imports above third party ones

glide.yaml
```diff
@@ -0,0 +1,66 @@
...
+  version: f2499483f923065a842d38eb4c7f1927e6fc6e6d
+- package: github.com/gorilla/mux
+  version: 18fca31550181693b3a834a15b74b564b3605876
+testImport:
+- package: github.com/stretchr/testify
+  version: 6fe211e493929a8aac0469b93f28b1d0688a9a3a
+- package: github.com/pmezard/go-difflib
+  version: d8ed2627bdf02c080bf22230dbb337003b7aba2d
+- package: github.com/davecgh/go-spew
+  version: 5215b55f46b2b919f50a1df0eaa5886afe4e3b3d
```
prateek, line 66: nit: insert newline at the end

handler/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+	"time"
+
+	"github.com/m3db/m3ctl/handler"
+	"github.com/m3db/m3x/instrument"
+
+	"github.com/gorilla/mux"
+)
+
+const (
+	ok          = "OK"
```
prateek: maybe make this a type to make it explicit -

```
type HealthStatus string

const (
    Ok HealthStatus = "OK"
    Fail            = "FAIL"
)
```

handler/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+	fail        = "FAIL"
+	healthURL   = "/health"
+	unknownName = "unkown"
+)
+
+type healthCheckResult struct {
```
prateek: Does it make sense for this to be exported so other packages can re-use?

handler/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+type controller struct {
+	iOpts instrument.Options
+}
+
+// NewController creates a new rules controller
+func NewController(iOpts instrument.Options) handler.Controller {
+	return &controller{iOpts: iOpts}
+}
+
+func healthCheck() string {
```
prateek: Why define this function instead of directly using the variable?

dgromov: This is the healthCheck in the skeleton implementation. It will be more interesting in the future.

handler/health/health.go
dgromov: Potentially but outside of the scope of this commit. It should go into m3x at that point though.

PR #3, opened 2017-06-21
Port can now be suplied by env var

dgromov:
Things running in docker like to pass in ports to listen on. Adding this option.


services/m3ctl/main/main.go
```diff
@@ -100,6 +108,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	envPort := os.Getenv(portEnvVar)
+
+	if envPort != "" {
+		if p, err := strconv.Atoi(envPort); err == nil {
+			cfg.HTTP.Port = p
```
prateek: nit: log a message indicating you're overriding the config value with an env var

services/m3ctl/main/main.go
```diff
@@ -100,6 +108,20 @@ func main() {
...
+		} else {
+			logger.Fatalf("%s (%s) is not a valid port number", envPort, portEnvVar)
+		}
+	} else {
```
prateek: This check should be outside the else, i.e.

```
if envPort != "" {
  ... 
}

if cfg.HTTP.Port == 0 {
   ...
}
```

services/m3ctl/main/main.go
```diff
@@ -47,7 +55,7 @@ type serverConfig struct {
 }
 
 // NewHTTPServerOptions create a new set of http server options.
```
prateek: nit: `s/New/new`

prateek (approved):
LGTM

PR #4, opened 2017-08-10
Swagger api spec for r2 api

dgromov:
Adding swagger api spec and route to serve it.

Also updated some of the existing routes to match it just for testing.


handler/swagger/swagger.go
```diff
@@ -0,0 +1,55 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swagger
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/m3db/m3ctl/handler"
```
xichen2020: Convention is to put `github.com/m3db/...` imports before 3rd party imports, with a blank line in between to separate them.

handler/swagger/swagger.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/m3db/m3x/instrument"
+)
+
+const (
+	swaggerURL = "/rules/swagger/"
```
xichen2020: Should this be `/m3ctl/swagger/`? there's nothing that's rule-specific here

dgromov: I'm more ok with just  ```/swagger/ ```. Its a global thing that will serve both r2api and m3ctl stuff.

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "description": "R2 client API",
+    "version": "0.0.1",
```
prateek: nit: use this instead of the version embedded in the `basePath`

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
...
+    },
+    {
+      "name": "rollup-rule",
+      "description": "Operations on rollup rules"
+    }
+  ],
+  "schemes": [
+    "http"
+  ],
+  "paths": {
```
prateek, line 35: maybe add [examples](https://swagger.io/docs/specification/adding-examples/) too?

handler/swagger/swagger.go
xichen2020: Right, but will you have both `/swagger/r2/v1/api` and `/swagger/m3ctl/v1/api`? I was thinking you can have `/m3ctl/swagger/r2/v1/api` for R2 stuff, and then in the future you can have `/m3ctl/swagger/m3admin/v1/api` etc.

public/swagger/swagger.json
dgromov: When this spec is more finalized, sure.

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
...
+        },
+        "rollupRules": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/RollupRule"
+          }
+        }
+      }
+    },
+    "MappingRule": {
```
saminzadeh, line 764: Is there a more descriptive name we could use here? MappingRule sounds generic and vague to me. 
Ideas ?
- "RetentionRule"
- "RetentionMappingRule"

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
...
+          "type": "string",
+          "format": "uuid"
+        },
+        "name": {
+          "type": "string"
+        },
+        "filters": {
+          "type": "string"
+        },
+        "targets": {
```
saminzadeh: Are targets tag names?

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "cutoverTime": {
```
saminzadeh: What time format are we supporting here?

public/swagger/swagger.json
```diff
@@ -0,0 +1,977 @@
...
+          "type": "string",
+          "format": "uuid"
+        },
+        "name": {
+          "type": "string"
+        },
+        "filters": {
+          "type": "string"
+        },
+        "policies": {
```
saminzadeh: Maybe a bit vague here too? As reference, Carbon (Graphite) names this retention.

Given the following format: **"10s:7d"**
**retention** = "10s:7d"
**frequency** = "10s"
**history** = "7d"

public/swagger/swagger.json
```diff
@@ -0,0 +1,997 @@
...
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "string",
+          "format": "uuid"
+        },
+        "name": {
+          "type": "string"
+        },
+        "filters": {
```
saminzadeh: This should be singular "filter" since its string rather than array.

public/swagger/swagger.json
```diff
@@ -0,0 +1,997 @@
...
+          }
+        },
+        "cutoverNanos": {
+          "type": "integer",
+          "description": "unixNanos"
+        },
+        "lastUpdateBy": {
+          "type": "string"
+        },
+        "lastUpdatedNanos": {
```
saminzadeh: Is it important to have the timestamps in nanos instead of milliseconds? In other words, do we need this much precision?

Some clients don't have great support for nanoseconds (eg Javascript)

public/swagger/swagger.json
saminzadeh: Also at some point, I would add what a valid "filter" and "policy" is in the top of the docs.

saminzadeh (approved):
A few small comments but REST API changes LGTM.

PR #6, opened 2017-08-30
R2api skeleton

dgromov:


services/r2ctl/server/server.go
```diff
@@ -24,18 +24,23 @@ import (
...
-	if err := registerHandlers(router, opts.InstrumentOptions()); err != nil {
-		return nil, err
+func registerHandlers(router *mux.Router, controllers []handler.Controller) {
+	for _, c := range controllers {
+		c.RegisterHandlers(router)
 	}
+}
+
+// NewServer creates a new http server for R2
+func NewServer(address string, serverOpts Options, controllers []handler.Controller) *http.Server {
```
prateek: I'm missing some context, could you clarify the difference between `controller`, `handler`, and `operator`.

handler/r2/api.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// API holds all the state for a multi backend api  instance.
+type api struct {
+	iOpts    instrument.Options
+	operator r2.Operator
+}
+
+// API exposes the http functions available to any implementation of an r2api.
+type API interface {
```
prateek: Wondering why this is an interface instead of a straight struct, would there ever be another implementation of the `API`?

dgromov: Might be useful as a test fixture? If only to test the register handlers call?

services/r2ctl/server/server.go
dgromov: Handler is just logical grouping of routes. The directory can just as easily be called Controllers.

A Controller is a wrapper around a set of http handlers that registers them with the server and provides some useful state. State can be instrumentation or a path prefix to expose them under. In the case of r2, it also has an implementation of an API to call the functions on. 

An Operator is the thing that actually performs the operations for r2. Each r2 api instance has an operator associated with it. For starters, I'm implementing a stub operator that won't actually talk to etcd but instead will return dummy results. When I implement the real Operator we'll just have to swap out the stub for that one.

prateek: talked offline, some renaming:

- `API` -> `Handlers`
- `Operator` -> `Service`
- `Controller` -> `Router`
- `Handler` -> folder goes away, separate folders for each router

r2/r2.go
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
prateek: nit: rename file to `router.go`

services/r2ctl/main/main.go
```diff
@@ -29,15 +29,20 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/m3db/m3ctl/health"
+	"github.com/m3db/m3ctl/r2"
+	"github.com/m3db/m3ctl/r2/kv"
+	"github.com/m3db/m3ctl/services/r2ctl/server"
+	"github.com/m3db/m3ctl/swagger"
+
```
prateek: nit: skip this new line or make the other files follow this convention too (e.g. `r2/r2.go` groups `m3db/...` org and package imports)

r2/handlers.go
```diff
@@ -0,0 +1,149 @@
...
+package r2
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/m3db/m3x/instrument"
+)
+
+// API holds all the state for a multi backend api  instance.
```
prateek: nit: comment needs to be updated

health/health.go
```diff
@@ -48,13 +48,13 @@ type healthCheckResult struct {
...
 
-type controller struct {
+type router struct {
 	iOpts instrument.Options
 }
 
-// NewController creates a new rules controller
-func NewController(iOpts instrument.Options) handler.Controller {
-	return &controller{iOpts: iOpts}
+// NewRouter creates a new rules controller
```
prateek: rename to `Routes` or some such

prateek:
More offline chats:
- Merging `router`+ `handler` => `Service` (it has a `Register(Endpoints|Handlers)` method
- `Service`/`Operator` renamed to `Store`/`Storage`...

r2/service.go
```diff
@@ -0,0 +1,172 @@
...
+// THE SOFTWARE.
+
+package r2
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/m3db/m3ctl/services/r2ctl/server"
+
```
prateek: import nit

prateek (approved):
LGTM apart from import orderings

PR #8, opened 2017-09-14
[glide] Update m3x, m3cluster, and m3metrics dependencies

jeromefroe:
This PR updates the m3x, m3cluster, and m3metrics dependencies and fixes any imports accordingly.


jeromefroe:
Need to update m3metrics and then this repo actually.

PR #9, opened 2017-09-22
[WIP] R2Api http implementation + Stub backing store

dgromov:
This includes the http handlers that will be able to serve the r2api against some backing store.
It is wired to a stubbed out backing store that is also implemented here.


README.md
```diff
@@ -15,3 +15,19 @@ This project is released under the [MIT License](LICENSE.md).
 [ci]: https://travis-ci.org/m3db/m3ctl
 [cov-img]: https://coveralls.io/repos/m3db/m3ctl/badge.svg?branch=master&service=github
 [cov]: https://coveralls.io/github/m3db/m3ctl?branch=master
+
+### Run the R2API
+
+```bash
+git clone --recursive https://github.com/m3db/m3ctl.git
+cd m3ctl
+glide install
```
prateek: nit: i suggest `glide install -v`

services/r2ctl/main/main.go
```diff
@@ -145,8 +136,7 @@ func main() {
 
 	services := []server.Service{
 		health.NewService(instrumentOpts),
-		swagger.NewService(instrumentOpts, "r2/v1"),
-		r2.NewService(instrumentOpts, "r2/v1", kv.NewStore(kv.StoreConfig{})),
+		r2.NewService(instrumentOpts, "r2/v1", stub.NewStore(instrumentOpts)),
```
prateek: do you mean to drop swagger here?

services/r2ctl/main/main.go
```diff
@@ -126,17 +117,17 @@ func main() {
 		logger.Fatalf("No valid port configured. Can't start.")
 	}
 
-	scope, closer, err := cfg.Metrics.NewRootScope()
-	if err != nil {
-		logger.Fatalf("error creating metrics root scope: %v", err)
-	}
-	defer closer.Close()
+	// scope, closer, err := cfg.Metrics.NewRootScope()
```
prateek: assuming this is for testing and you'll undo before merge, yea?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package r2
```
prateek: general nit: try to treat `types.go` as a header file, rename to something else if you're going to leave as is.

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+
+func (p payload) write(w http.ResponseWriter) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(p.status)
+	w.Write(p.data)
+}
+
+func writeAPIResponse(w http.ResponseWriter, code int, msg string) {
+	j, _ := json.Marshal(apiResponse{Code: code, Message: msg})
+	newPayload(j, code).write(w)
```
cw9: so code is stored in payload and payload.data, is this necessary?

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+}
+
+func (p payload) write(w http.ResponseWriter) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(p.status)
+	w.Write(p.data)
+}
+
+func writeAPIResponse(w http.ResponseWriter, code int, msg string) {
+	j, _ := json.Marshal(apiResponse{Code: code, Message: msg})
```
prateek: nit: don't drop err here. can do a `panic` if you're sure this should never fail

r2/service.go
prateek: nit: check the err here too

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+// TODO(dgromov): Make this return a list of validation errors
+func parseRequest(s interface{}, body io.ReadCloser) string {
+	if err := json.NewDecoder(body).Decode(s); err != nil {
+		return fmt.Sprintf("Malformed Json: %s", err.Error())
+	}
+
+	validate := validator.New()
+
+	validate.RegisterTagNameFunc(func(fld reflect.StructField) string {
+		name := strings.SplitN(fld.Tag.Get("json"), ",", 2)[0]
```
prateek: are you guaranteed `strings.SplitN(fld.Tag.Get("json"), ",", 2)` always has len > 0

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+	return ""
+}
+
 func (s *service) readNamespaces(w http.ResponseWriter, _ *http.Request) {
-	w.WriteHeader(http.StatusNotImplemented)
-	fmt.Fprintf(w, "Placeholder for readNamespace")
+	names, v, err := s.store.FetchNamespaces()
+	if err != nil {
+		writeAPIResponse(w, http.StatusInternalServerError, "Could not read namespaces")
```
prateek: nit: check if the write fails and log if so. will make debugging a lot easier.

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+		nss[i] = namespaceJSON{ID: n}
+	}
+
+	j, err := json.Marshal(namespacesJSON{
+		Version:    v,
+		Namespaces: nss,
+	})
+
+	if err != nil {
+		writeAPIResponse(w, http.StatusInternalServerError, "Could not create response object")
```
prateek: same as above, log if writeAPIResponse fails. Need to do this in all function below too

r2/store.go
```diff
@@ -20,5 +20,36 @@
...
 package r2
 
-// Store is a construct that performs operations against a backing rule store.
-type Store interface{}
+import "github.com/m3db/m3metrics/rules"
+
+// Store is a construct that can perform operations against a backing rule store.
+type Store interface {
+	// Fetch namespaces and their version
+	FetchNamespaces() ([]string, int, error)
```
prateek, line 28: minor nit: your utility functions in `service.go` go by `readNamespaces` and so on, here you're using `FetchNamespaces`. Pick one to stay consistent

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+
+	newPayload(j, http.StatusCreated).write(w)
 }
 
-func (s *service) updateRollupRule(w http.ResponseWriter, _ *http.Request) {
-	w.WriteHeader(http.StatusNotImplemented)
-	fmt.Fprintf(w, "Placeholder for updateRollupRule")
+func (s *service) updateRollupRule(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
```
prateek: you seem to have a lot of similar arg checking in place, might be useful to write a utility helper - 

e.g. 

```
func parseVars(r *http.Request, names ...string) (map[string]string, error) {
   vars := mux.Vars(r)
   for _, n := range names {
     if _, ok := vars[n]; !ok {
       return nil, fmt.Errorf("Could not parse %s", n)
     }
   }
   return vars, nil
}

func (s *service) updateRollupRule(w http.ResponseWriter, r *http.Request) {
    vars, err := parseVars(r, "namespaceID", "ruleID")
    if err != nil {
       writeAPIResponse(w, http.StatusBadRequest, err.Error())
    }
    // ... 
} 
```

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+}
+
+// TODO(dgromov): Make this return a list of validation errors
+func parseRequest(s interface{}, body io.ReadCloser) string {
+	if err := json.NewDecoder(body).Decode(s); err != nil {
+		return fmt.Sprintf("Malformed Json: %s", err.Error())
+	}
+
+	validate := validator.New()
```
prateek: Is there a reason you have to do this validation manually (instead of relying on the validator/unmarshall interface)? if so, put a brief note about it

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	var n namespaceJSON
+	errs := parseRequest(&n, r.Body)
+	if len(errs) > 0 {
+		writeAPIResponse(w, http.StatusBadRequest, errs)
+		return
+	}
+
+	id, err := s.store.CreateNamespace(n.ID)
+	if err == nil {
+		writeAPIResponse(w, http.StatusCreated, fmt.Sprintf("Created namespace %s", id))
```
cw9: return here to save the else case? Also any particular reason to handle err == nil case first here?

services/r2ctl/main/main.go
```diff
@@ -72,26 +70,19 @@ func (c *serverConfig) newHTTPServerOptions(
 	return opts
 }
 
-type kvClientConfiguration struct {
-	Etcd *etcd.Configuration `yaml:"etcd"`
-}
-
 type configuration struct {
-	// KVStore Config.
-	KVClient kvClientConfiguration `yaml:"kvClient" validate:"nonzero"`
```
prateek, line 81: dropping this for testing?

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+
+	validate := validator.New()
+
+	validate.RegisterTagNameFunc(func(fld reflect.StructField) string {
+		name := strings.SplitN(fld.Tag.Get("json"), ",", 2)[0]
+		return name
+	})
+
+	required := make([]string, 0)
```
prateek: why not just `var required []string` ?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/m3db/m3metrics/policy"
+	"github.com/m3db/m3metrics/rules"
+)
+
+const (
+	tagFilterListSeparator = " "
```
prateek: Is there ever a chance this needs to be changed by the user?

prateek: If not, might still be useful to have these exported (and a comment describing what they represent)

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+
+// nolint
+func newMappingRules(mrs rules.MappingRules) []*rules.MappingRuleView {
+	result := make([]*rules.MappingRuleView, len(mrs))
+	idx := 0
+	for _, m := range mrs {
+		if len(m) == 0 {
+			continue
+		}
+		result[idx] = m[len(m)-1]
```
prateek: nit: expose a function on the type `m` to indicate what `m[len(m)-1]` is supposed to be (presumably something like `ActiveRule` or `LatestRule`)

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+
+// nolint
+func newRollupRules(rrs rules.RollupRules) []*rules.RollupRuleView {
+	result := make([]*rules.RollupRuleView, len(rrs))
+	idx := 0
+	for _, r := range rrs {
+		if len(r) == 0 {
+			continue
+		}
+		result[idx] = r[len(r)-1]
```
prateek: same as `newMappingRules` feedback here too

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	Version    int             `json:"version"`
+	Namespaces []namespaceJSON `json:"namespaces"`
+}
+
+type filters map[string]string
+
+func (f filters) MarshalJSON() ([]byte, error) {
+	tmp := make([]string, 0, len(f))
+	for k, v := range f {
+		tmp = append(tmp, fmt.Sprintf("%s:%s", k, v))
```
prateek: do you mean to use your constants here, and in the line below?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+
+// Namespace Level
+type namespacesJSON struct {
+	Version    int             `json:"version"`
+	Namespaces []namespaceJSON `json:"namespaces"`
+}
+
+type filters map[string]string
+
+func (f filters) MarshalJSON() ([]byte, error) {
```
prateek: curious: does golang not marshall/unmarshall `map[string]string` natively?

r2/service.go
```diff
@@ -59,86 +64,480 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+
+	mappingRuleID, exists := vars["ruleID"]
+	if !exists {
+		writeAPIResponse(w, http.StatusBadRequest, "Could not parse mappingRuleID")
+		return
+	}
+
+	err := s.store.DeleteMappingRule(namespaceID, mappingRuleID)
+	if err == nil {
```
cw9: nit: if err := s.store.DeleteMappingRule(namespaceID, mappingRuleID); err != nil {...}

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+		if len(parts) != 2 {
+			return NewBadInputError(fmt.Sprintf("invalid tag filter pair %s", p))
+		}
+		(*f)[parts[0]] = parts[1]
+	}
+	return nil
+}
+
+type mappingRuleJSON struct {
+	ID             string          `json:"id,omitempty"`
```
prateek: typo in your json tags, should be `json:"id" validate:"omitempty"`

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	Name           string          `json:"name" validate:"required"`
+	CutoverMillies int64           `json:"cutoverMillies,omitempty"`
```
prateek: typo in your json tags, should be `json:"cutoverMillis" validate:"omitempty"`

(note: not `..Millies`)

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+func newRollupTargetJSON(t rules.RollupTargetView) rollupTargetJSON {
+	return rollupTargetJSON{
+		Name:     t.Name,
+		Tags:     t.Tags,
+		Policies: t.Policies,
+	}
+}
+
+type rollupRuleJSON struct {
+	ID             string             `json:"id,omitempty"`
```
prateek: typo in your json tags, should be `json:"id" validate:"omitempty"`

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	Name           string             `json:"name" validate:"required"`
+	Filters        filters            `json:"filter" validate:"required"`
+	Targets        []rollupTargetJSON `json:"targets" validate:"required,dive,required"`
+	CutoverMillies int64              `json:"cutoverMillies,omitempty"`
```
prateek: typo in your json tags, should be `json:"cutoverMillies" validate:"omitempty"`

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	targets := make([]rollupTargetJSON, len(rrv.Targets))
+	for i, t := range rrv.Targets {
+		targets[i] = newRollupTargetJSON(t)
+	}
+	return rollupRuleJSON{
+		ID:             rrv.ID,
+		Name:           rrv.Name,
+		Filters:        rrv.Filters,
+		Targets:        targets,
+		CutoverMillies: nanoToMilli(rrv.CutoverNanos),
```
prateek: why `Millies`?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+
+type rollupRuleHistory struct { // nolint
+	States []rollupRuleJSON `json:"states"`
+}
+
+func nanoToMilli(nanos int64) int64 {
+	return nanos / 1000000
+}
+
+// ConflictError represents either a version mismatch writing data or a data conflict issue.
```
prateek: Do you mean for these to be exported types?

r2/store.go
```diff
@@ -20,5 +20,36 @@
...
+
+	CreateNamespace(namespaceID string) (string, error)
+
+	DeleteNamespace(namespaceID string) error
+
+	FetchRuleSet(namespaceID string) (*CurrentRuleSet, error)
+
+	FetchMappingRule(namespaceID, mappingRuleID string) (*rules.MappingRuleView, error)
+
+	CreateMappingRule(namespaceID string, mrv rules.MappingRuleView) (*rules.MappingRuleView, error)
```
cw9: any reason for the input to be a view struct and return a pointer?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+		Targets: targets,
+	}
+}
+
+type rollupRuleHistory struct { // nolint
+	States []rollupRuleJSON `json:"states"`
+}
+
+func nanoToMilli(nanos int64) int64 {
+	return nanos / 1000000
```
prateek: could you use `m3x/time.FromNormalizedTime` instead

r2/service.go
prateek: might be worth it to enable the `errCheck` linter using gometalinter.

r2/stub/store.go
```diff
@@ -0,0 +1,526 @@
...
+	data  *stubData
+	iOpts instrument.Options
+}
+
+// NewStore creates a new stub
+func NewStore(iOpts instrument.Options) r2.Store {
+	return &store{data: &dummyData, iOpts: iOpts}
+}
+
+// FetchNamespaces ...
```
cw9: don't think it will complain if you don't have this comment

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	currentMappingRules := newMappingRules(mrs)
+	currentRollupRules := newRollupRules(rrs)
+
+	return &CurrentRuleSet{
+		Namespace:    string(r.Namespace()),
+		Version:      r.Version(),
+		CutoverNanos: r.CutoverNanos(),
+		MappingRules: currentMappingRules,
```
cw9: nit: directly do MappingRules: newMappingRules(mrs)

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	result := make([]*rules.MappingRuleView, len(mrs))
+	idx := 0
+	for _, m := range mrs {
+		if len(m) == 0 {
+			continue
+		}
+		result[idx] = m[len(m)-1]
+		idx++
+	}
+	return result
```
cw9: should return result[:idx]?  is there a test for this?

Is it simpler if you just use append in this case?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	result := make([]*rules.RollupRuleView, len(rrs))
+	idx := 0
+	for _, r := range rrs {
+		if len(r) == 0 {
+			continue
+		}
+		result[idx] = r[len(r)-1]
+		idx++
+	}
+	return result
```
cw9: same, use append or result[:idx]

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (f *filters) UnmarshalJSON(data []byte) error {
+	str := string(data)
+	unquoted, err := strconv.Unquote(str)
+	if err != nil {
+		return err
+	}
+
+	*f = make(map[string]string)
```
cw9: nit: I'd create a temp map here, and assign *f to it before the final return in case things error out and f end up in a weird state.

r2/types.go
cw9: use tagFilterSeparator instead of ":"?

services/r2ctl/main/main.go
```diff
@@ -72,26 +70,19 @@ func (c *serverConfig) newHTTPServerOptions(
...
-
 	// Logging configuration.
 	Logging log.Configuration `yaml:"logging"`
 
 	// Metrics configuration.
-	Metrics instrument.MetricsConfiguration `yaml:"metrics"`
+	// Metrics instrument.MetricsConfiguration `yaml:"metrics"`
```
cw9: remove the "//"?

services/r2ctl/main/main.go
cw9: do we need this?

services/r2ctl/main/main.go
dgromov: Yeah, swagger is served as a static file out of /public. It doesn't need a service.

services/r2ctl/main/main.go
dgromov: Yep.

r2/types.go
dgromov: Some ideas:

- Move the whole thing into store.go? 
- Leave types in types.go and move implementations into store.go? 
- Call this something like common.go?

r2/service.go
dgromov: will do

r2/service.go
dgromov: good point

r2/service.go
dgromov: can do

services/r2ctl/main/main.go
dgromov: yea

r2/types.go
dgromov: It does but a user using this inputs a string like "tag1:value1 tag2:value2". I'm marshaling a map[string]string into a string and back. The alternative is to have another intermediate type just to receive input and then something like this function to make a the struct I actually need.

r2/types.go
dgromov: Not according to: https://golang.org/pkg/encoding/json/#Marshal. 

omitempty is not a validation step.

r2/types.go
dgromov: Browsers don't have  64 bit ints. Javascript would get confused if I returned a timestamp in nanos.

r2/types.go
dgromov: Yeah, I'm creating them in the implementations of the store interface.

r2/types.go
dgromov: If you're asking why I added the e... Read that millies was more correct somewhere and went with it.

r2/types.go
dgromov: Sure

services/r2ctl/main/main.go
prateek: cool

services/r2ctl/main/main.go
prateek: Cool

r2/types.go
prateek: (2) sounds good

services/r2ctl/main/main.go
prateek: cool

r2/types.go
prateek: cool

r2/types.go
prateek: Yeah I was commenting on the spelling. It's better to stick to popular convention, even if grammatically incorrect for APIs. Change to `Millis`.

r2/types.go
prateek: Is the idea for the API user to be able to distinguish different types of failures?

r2/kv/store.go
```diff
@@ -38,3 +38,59 @@ type StoreConfig struct {
 func NewStore(cfg StoreConfig) r2.Store {
 	return &store{cfg: cfg}
 }
+
+func (s *store) FetchNamespaces() ([]string, int, error) {
+	return nil, 0, nil
```
prateek, line 43: take/leave: (just a personal preference when stubbing) I like to `panic("not implemented")` instead of returning a default value. Makes it easier to track down issues as you're filling in code

r2/types.go
prateek: Why can't you represent the values as `map[string]string`, and expect the users to follow that convention?

r2/types.go
```diff
@@ -0,0 +1,308 @@
...
+	ID string `json:"id" validate:"required"`
+}
+
+// Namespace Level
+type namespacesJSON struct {
+	Version    int             `json:"version"`
+	Namespaces []namespaceJSON `json:"namespaces"`
+}
+
+type filters map[string]string
```
prateek: For `filters` to be used in your `...View` types, don't they need to be exported?

r2/types.go
dgromov: Its basically the same as writing a m3ql. Its nice to have a consistent user exp.

r2/types.go
dgromov: More for someone implementing a store to know how to map http codes to errors.

r2/types.go
dgromov: filters is just an alias in this package so that I can define things on it. Views have a field called filters that is also a map[string]string

r2/types.go
cw9: It's pretty weird to have the implementations not in the same file with the Struct definition, usually we only leave interfaces in types.go, structs and implementations should go to separate files

r2/service.go
dgromov: I don't actually need this validation. I'm only checking if its set to something. If it isn't set to anything the regex for the route matching wouldn't work. Can just get rid of it.

services/r2ctl/server/service.go
```diff
@@ -26,4 +26,7 @@ import "github.com/gorilla/mux"
 type Service interface {
 	// RegisterHandlers wires the http handlers for this Service with the given router.
 	RegisterHandlers(*mux.Router)
+
+	// Root returns the root of this service
+	Root() string
```
prateek: `Root` seems weird, could you give this a better name? `RelativeURLPath` or something

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+}
+
+func newPayload(d []byte, status int) payload {
+	return payload{data: d, status: status}
+}
+
+func (p payload) write(w http.ResponseWriter) error {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(p.status)
+	if _, err := w.Write(p.data); err != nil {
```
prateek: nit: this can be simplified to

```
// w.Header()..
// w.WriteHeader(...)
_, err := w.Write(p.data)
return err
```

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	}
+	return ""
+}
+
+func writeAPIResponse(w http.ResponseWriter, code int, msg string) error {
+	j, err := json.Marshal(apiResponse{Code: code, Message: msg})
+	if err != nil {
+		return err
+	}
+	if err := newPayload(j, code).write(w); err != nil {
```
prateek: nit: could you break the var chaining, the following feels more idiomatic go

```
payload := newPayload(j, code)
return payload.write(w)
```

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+		err = writeAPIResponse(w, http.StatusConflict, opError.Error())
+	case NotFoundError:
+		err = writeAPIResponse(w, http.StatusNotFound, opError.Error())
+	default:
+		err = writeAPIResponse(w, http.StatusInternalServerError, opError.Error())
+	}
+
+	// Getting here means that the error handling failed. Trying to convey what was supposed to happen.
+	if err != nil {
+		http.Error(
```
prateek: circular failure mode here ~ imagine if the http response writer was failing, your reaction here is to try again. maybe add a log of the error too

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
 	return &service{iOpts: iOpts, store: store, rootPrefix: rootPrefix}
 }
 
-func (s *service) readNamespaces(w http.ResponseWriter, _ *http.Request) {
-	w.WriteHeader(http.StatusNotImplemented)
-	fmt.Fprintf(w, "Placeholder for readNamespace")
+type payload struct {
```
prateek: I don't understand the use of this type. Why do you need to indicate the status code here? Won't the HTTP layer address that?

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	nss := make([]namespaceJSON, len(names))
+	for i, n := range names {
+		nss[i] = namespaceJSON{ID: n}
+	}
+	return s.write(w, namespacesJSON{Version: v, Namespaces: nss})
+}
+
+func (s *service) fetchNamespace(w http.ResponseWriter, r *http.Request) error {
+	vars := mux.Vars(r)
+
```
prateek: nit: redundant newline

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(p.status)
+	if _, err := w.Write(p.data); err != nil {
+		return fmt.Errorf("cannot write http payload. %s", err.Error())
+	}
+	return nil
+}
+
+// TODO(dgromov): Make this return a list of validation errors
+func parseRequest(s interface{}, body io.ReadCloser) string {
```
prateek: nit: make this return something that implements `Error() string`

public/index.html
```diff
@@ -0,0 +1 @@
+Weclome to M3CTL.
```
prateek: typo: `Welcome`

r2/service.go
```diff
@@ -59,113 +78,569 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	return rules.RollupRuleView{
+		ID:      r.ID,
+		Name:    r.Name,
+		Filters: r.Filters,
+		Targets: targets,
+	}
+}
+
+type rollupRuleHistoryJSON struct {
+	States []rollupRuleJSON `json:"rollupRules"`
```
prateek: do you mean to reconcile the difference between var and type name here?

public/index.html
dgromov: lol

prateek: :)

public/r2/v1/swagger/swagger.json
```diff
@@ -14,18 +14,14 @@
     "tags": [
         {
             "name": "namespaces",
-            "description": "Operations at the namespaces level"
+            "description": "Operations at the namespace level"
```
prateek, line 17: per our offline chat, look into golang swagger annotation based generator

dgromov: Definitely worth doing. Will make an issue on this project.

dgromov: https://github.com/m3db/m3ctl/issues/11

services/r2ctl/server/service.go
```diff
@@ -26,4 +26,7 @@ import "github.com/gorilla/mux"
 type Service interface {
 	// RegisterHandlers wires the http handlers for this Service with the given router.
 	RegisterHandlers(*mux.Router)
+
+	// URLPrefix returns the prefix for of this service
```
prateek: typo: `... for of this service`

prateek (approved):
LGTM

PR #10, opened 2017-09-22
Simple Auth

dgromov:


services/r2ctl/server/auth.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	// UserIDField is the name of the variable that simple auth will put in the mux context
+	UserIDField = "userID"
+)
+
+// HTTPAuthService defines how to handle requests for various http authentication methods.
+type HTTPAuthService interface {
+	// User gets some identifying information about the user from the http request.
+	User(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc)
```
prateek: does the signature have to be this for chaining?

r2/service.go
```diff
@@ -59,14 +60,17 @@ func NewService(iOpts instrument.Options, rootPrefix string, store Store) server
...
+	id := context.Get(r, "userID")
+	fmt.Fprintf(w, "Placeholder for readNamespace - User: %s", id)
 }
 
-func (s *service) readNamespace(w http.ResponseWriter, _ *http.Request) {
+func (s *service) readNamespace(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNotImplemented)
-	fmt.Fprintf(w, "Placeholder for readNamespace")
+	id := context.Get(r, "userID")
+	// fmt.Println("%v", context.GetAll(r))
```
cw9: remove this?

services/r2ctl/server/auth.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// User looks for a header defining a user name. If it can't find it, spits back not authorized.
+func (a SimpleAuth) User(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	userID := r.Header.Get(a.userNameHeader)
+
+	if userID == "" {
+		http.Error(rw, fmt.Sprintf("No %s Provided", a.userNameHeader), 401)
+	} else {
+		// http.Error(rw, fmt.Sprintf("%s Provided", userID), 401)
```
cw9: remove this

services/r2ctl/server/auth.go
dgromov: Yeah. https://github.com/urfave/negroni#handlers

auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+// THE SOFTWARE.
+
+package auth
+
+import (
+	"context"
+	"net/http"
+)
+
+type key int
```
xichen2020: Doesnt look like this is used? Also perhaps better to call it `Type`?

auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+
+const (
+	// UserIDField is a key
+	UserIDField key = iota
+)
+
+// HTTPAuthService defines how to handle requests for various http authentication methods.
```
xichen2020: I think the convention is to capitalize `http` in comments to `HTTP`.

auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+type HTTPAuthService interface {
+	// Authenticate should return a handler that performs some check on the request coming into the given handler
+	// and then runs the handler if it is. If the request is authenticated successfully, it should call SetUser
+	// to make the callers id available to the service in a global context.
+	Authenticate(next http.Handler) http.Handler
```
xichen2020: `NewAuthHandler` might be better? I'd expect a method named `Authenticate` to do the auth work instead of returning a handler.

auth/simple.go
```diff
@@ -0,0 +1,72 @@
...
+
+package auth
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// SimpleAuthConfig holds this configuration necessary for a simple auth implementation
```
xichen2020: Comments should end with a period. Same below.

auth/simple.go
```diff
@@ -0,0 +1,72 @@
...
+type SimpleAuthConfig struct {
+	// This is an http header that identifies the user performing the operation
+	UserNameHeader string `yaml:"userNameHeader" validate:"nonzero"`
```
xichen2020: `UserIDHeader`?

auth/simple.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// SimpleAuth is an naive authentication method that just checks for a HTTPHeader identifying the user
+// performing the action.
+type SimpleAuth struct {
```
xichen2020: This probably doesn't need to be exported?

auth/simple.go
```diff
@@ -0,0 +1,72 @@
...
+	userNameHeader string
+}
+
+// NewSimpleAuth creates a new simple auth instance given using the provided config.
+func (ac SimpleAuthConfig) NewSimpleAuth() SimpleAuth {
```
xichen2020: Convention is to return the interface type here.

auth/simple.go
```diff
@@ -0,0 +1,72 @@
...
+			return
+		}
+		ctx := a.SetUser(r.Context(), userID)
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// SetUser sets the user making the changes to the api.
+func (a SimpleAuth) SetUser(parent context.Context, userID string) context.Context {
+	return context.WithValue(parent, UserIDField, userID)
```
xichen2020: Make `UserIDField` part of the config instead of a const?

config/base.yaml
```diff
@@ -8,3 +8,5 @@ http:
   readTimeout: 10s
   writeTimeout: 10s
 
+auth:
+  userNameHeader: user
```
xichen2020: nit: add a empty line?

r2/errors.go
```diff
@@ -60,3 +60,11 @@ type NotFoundError struct{ msg string }
 func NewNotFoundError(msg string) NotFoundError { return NotFoundError{msg: msg} }
 
 func (e NotFoundError) Error() string { return e.msg }
+
+// AuthError represents an error due to missing or invalid auth information.
+type AuthError struct{ msg string }
```
xichen2020: This doesn't need to be exported no? Can also do `type authError string`.

r2/errors.go
```diff
@@ -60,3 +60,11 @@ type NotFoundError struct{ msg string }
...
+
+// NewAuthError creates a new not found Error.
+func NewAuthError(msg string) AuthError { return AuthError{msg: msg} }
```
xichen2020: Convention is just to return `error`.

r2/errors.go
```diff
@@ -60,3 +60,11 @@ type NotFoundError struct{ msg string }
 func NewNotFoundError(msg string) NotFoundError { return NotFoundError{msg: msg} }
 
 func (e NotFoundError) Error() string { return e.msg }
+
+// AuthError represents an error due to missing or invalid auth information.
```
xichen2020: Should this be moved to the `auth` package instead? Doesn't seem like this is r2 specific.

r2/service.go
```diff
@@ -112,15 +113,21 @@ func writeAPIResponse(w http.ResponseWriter, code int, msg string) error {
 	return write(w, j, code)
 }
 
+type r2HandlerFunc func(http.ResponseWriter, *http.Request) error
```
xichen2020, line 116: Just use `http.HandlerFunc`?

services/r2ctl/main/main.go
```diff
@@ -76,6 +77,9 @@ type configuration struct {
 
 	// HTTP server configuration.
 	HTTP serverConfig `yaml:"http"`
+
+	// Simple Auth Config
```
xichen2020: nit: add a period at the end.

r2/service.go
```diff
@@ -112,15 +113,21 @@ func writeAPIResponse(w http.ResponseWriter, code int, msg string) error {
...
 	iOpts instrument.Options
-	fn    func(http.ResponseWriter, *http.Request) error
+	auth  auth.HTTPAuthService
 }
 
-func (h r2Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if err := h.fn(w, r); err != nil {
-		h.handleError(w, err)
-	}
+func (h r2Handler) handle(fn r2HandlerFunc) http.Handler {
```
xichen2020: nit: `wrap` might be a better name?

auth/auth.go
dgromov: it is used on line 32. Go's context.Context takes an interface{} as a key and go doesn't like it when you pass a basic type to that.

auth/simple.go
dgromov: this is the key in the context, not the header name. It shouldn't be configurable.

r2/errors.go
dgromov: 1. I need to initialize these things outside of these package. e.g. in r2/kv/store... So I do need to export them.

2. Good point. Can do that for all of the errors actually.

r2/service.go
dgromov: http.HandlerFunc doesn't return an error.

r2/service.go
dgromov: sure

auth/auth.go
xichen2020: Sure, are you going to call it `Type` for that matter?

auth/simple.go
xichen2020: Fair enough

r2/errors.go
xichen2020: Re 1: is exporting the constructor not sufficient if you just need to initialize it outside the package?

r2/service.go
xichen2020: Hmm yeah sounds good.

r2/errors.go
```diff
@@ -20,51 +20,48 @@
...
 
-// InternalError represents an unexpected server error.
-type InternalError struct{ msg string }
+// ConflictError represents either a version mismatch writing data or a data conflict issue.
+type conflictError string
 
-// NewInternalError creates a new Internal Error
-func NewInternalError(msg string) InternalError { return InternalError{msg: msg} }
+// NewConflictError creates a new Conflict Error
+func NewConflictError(msg string) error { return errors.New(msg) }
```
xichen2020: Should probably return `conflictError(msg)` here?

r2/errors.go
```diff
@@ -20,51 +20,48 @@
...
+func (e conflictError) Error() string { return string(e) }
 
 // VersionError represents a mismatch in the Namespaces or Ruleset version specified in the request
 // and the latest one.
-type VersionError struct{ msg string }
+type versionError string
 
 // NewVersionError creates a new Version Error
-func NewVersionError(msg string) VersionError { return VersionError{msg: msg} }
+func NewVersionError(msg string) error { return errors.New(msg) }
```
xichen2020: Should probably return `versionError(msg)` here?

r2/errors.go
```diff
@@ -20,51 +20,48 @@
...
-func (e VersionError) Error() string { return e.msg }
+func (e versionError) Error() string { return string(e) }
 
 // BadInputError represents an error due to malformed or invalid metrics.
-type BadInputError struct{ msg string }
+type badInputError string
 
 // NewBadInputError creates a new Bad Input Error.
-func NewBadInputError(msg string) BadInputError { return BadInputError{msg: msg} }
+func NewBadInputError(msg string) error { return errors.New(msg) }
```
xichen2020: Should probably return `badInputError(msg)` here?

r2/errors.go
```diff
@@ -20,51 +20,48 @@
...
-func (e BadInputError) Error() string { return e.msg }
+func (e badInputError) Error() string { return string(e) }
 
 // NotFoundError represents an error due to malformed or invalid metrics.
-type NotFoundError struct{ msg string }
+type notFoundError string
 
 // NewNotFoundError creates a new not found Error.
-func NewNotFoundError(msg string) NotFoundError { return NotFoundError{msg: msg} }
+func NewNotFoundError(msg string) error { return errors.New(msg) }
```
xichen2020: Should probably return `notFoundError(msg)` here?

r2/errors.go
```diff
@@ -20,51 +20,48 @@
...
-func (e NotFoundError) Error() string { return e.msg }
+func (e notFoundError) Error() string { return string(e) }
 
 // AuthError represents an error due to missing or invalid auth information.
-type AuthError struct{ msg string }
+type authError string
 
 // NewAuthError creates a new not found Error.
-func NewAuthError(msg string) AuthError { return AuthError{msg: msg} }
+func NewAuthError(msg string) error { return errors.New(msg) }
```
xichen2020: Should probably return `authError(msg)` here?

xichen2020 (approved):
LGTM

PR #13, opened 2017-10-09
Implementation against an etcd store

dgromov:


auth/noop.go
```diff
@@ -0,0 +1,52 @@
...
+
+package auth
+
+import (
+	"context"
+	"net/http"
+)
+
+const (
+	noopUser = "someone"
```
xichen2020: nit: might be better to simply have `noopUser` as the value here.

glide.lock
```diff
@@ -1,14 +1,24 @@
-hash: 4c7f11b17aaebaa2390b1a7fd1ff65b79e3ad13d759d071cadf96067ba39d0c8
-updated: 2017-09-28T16:52:13.647360205-04:00
+hash: 46a20cf3aceb51fa9c38f1d9d87791a7bc081f6a022260e0b458f2958913fe9f
+updated: 2017-10-09T17:13:15.718628298-04:00
 imports:
 - name: github.com/apache/thrift
   version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
   subpackages:
   - lib/go/thrift
 - name: github.com/beorn7/perks
```
xichen2020, line 8: Mind pinning the dependencies in `glide.yaml`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
-	UpdateHelper rules.RuleSetUpdateHelper
+// NewStore returns a new service that knows how to talk to a kv backed r2 store.
+func NewStore(iOpts instrument.Options, uh rules.RuleSetUpdateHelper, rs rules.Store) r2.Store {
+	return &store{iOpts: iOpts, updateHelper: uh, ruleStore: rs}
 }
 
-// NewStore returns a new service that knows how to talk to a kv backed r2 store.
-func NewStore(cfg StoreConfig) r2.Store {
-	return &store{cfg: cfg}
+func (s *store) updateMeta(author string) rules.UpdateMetadata {
```
xichen2020: nit: Take it or leave it, but I tend to keep public APIs at the top of the file and private methods at the bottom of the file to make it easy for readers.

xichen2020: nit: `updateMeta` -> `newUpdateMeta`.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	return s.updateHelper.NewUpdateMetadata(time.Now().UnixNano(), author)
```
xichen2020: nit: might wanna add a clock options to the store and use the `NowFn` on the clock options so it's easy to mock out time related logic in unit tests.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
 }
 
-func (s *store) FetchNamespaces() ([]string, int, error) {
-	return nil, 0, nil
+func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
+	nss, err := s.ruleStore.ReadNamespaces()
+	if err != nil {
+		return nil, nil
```
xichen2020: Should this be `return nil, err`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	}
+
+	namespaces, err := nss.NamespacesView()
+	if err != nil {
+		s.handleUpstreamError(err)
```
xichen2020: Do you want to return here?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	}
+
+	var liveNss []*rules.NamespaceView
```
xichen2020: nit: Allocate first with `liveNss := make([]*rules.NamespaceView, 0, namespaces.Namespaces)`

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	return &rules.NamespacesView{
+		Namespaces: liveNss,
+		Version:    namespaces.Version,
+	}, nil
 }
 
-func (s *store) CreateNamespace(namespaceID string) (string, error) {
-	return "", nil
+func (s *store) CreateNamespace(namespaceID string, author string) (*rules.NamespaceView, error) {
+	nss, err := s.ruleStore.ReadNamespaces()
```
xichen2020: What happens if the namespace doesn't exist? Do we return an error here?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+
+	return &rules.NamespacesView{
+		Namespaces: liveNss,
+		Version:    namespaces.Version,
+	}, nil
 }
 
-func (s *store) CreateNamespace(namespaceID string) (string, error) {
-	return "", nil
+func (s *store) CreateNamespace(namespaceID string, author string) (*rules.NamespaceView, error) {
```
xichen2020: Instead of `author`, can we pass an options struct with `author` in it? This way in the future if we need to add more stuff to the options, the API doesn't need to change.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+
+	rs := rules.NewEmptyRuleSet(namespaceID, s.updateMeta(author))
+
+	if revived {
+		rawRs, err := s.ruleStore.ReadRuleSet(namespaceID)
+		if err != nil {
+			return nil, s.handleUpstreamError(err)
+		}
+		rs = rawRs.ToMutableRuleSet().Clone()
+		if err = rs.Revive(s.updateMeta(author)); err != nil {
```
xichen2020: You already called `s.updateMeta(author)` above, perhaps reuse it?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+		}
+	}
+
+	if err = s.ruleStore.WriteAll(nss, rs); err != nil {
+		return nil, s.handleUpstreamError(err)
+	}
+
+	nss, err = s.ruleStore.ReadNamespaces()
+	if err != nil {
+		return nil, err
```
xichen2020: Why do we sometimes call `s.handleUpstreamError` and sometimes just return `err`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	}
+
+	ns, err := nss.Namespace(namespaceID)
```
xichen2020: Should this be handled from the UI side? Something like "on creation success, issue a fetch / refresh" request? This way the backend doesn't do manually refreshes after success.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	if err != nil {
+		return nil, r2.NewNotFoundError(fmt.Sprintf("namespace: %s does not exist", namespaceID))
+	}
+
+	view, err := ns.NamespaceView(0)
```
xichen2020: What's `0`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
 }
 
-func (s *store) DeleteNamespace(namespaceID string) error {
+func (s *store) DeleteNamespace(namespaceID string, author string) error {
+	nss, err := s.ruleStore.ReadNamespaces()
+	if err != nil {
+		return s.handleUpstreamError(err)
+	}
+	rs, err := s.ruleStore.ReadRuleSet(namespaceID)
+
```
xichen2020: nit: remove superfluous blank line.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
 	return nil
 }
 
 func (s *store) FetchRuleSet(namespaceID string) (*r2.CurrentRuleSet, error) {
-	return nil, nil
+	rs, err := s.ruleStore.ReadRuleSet(namespaceID)
+	if err != nil {
+		return nil, s.handleUpstreamError(err)
+	}
+	return r2.NewCurrentRuleSetView(rs)
```
xichen2020: Could probably just return `rs.Latest()` if you rebase off of master.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	if err != nil {
+		return nil, s.handleUpstreamError(err)
+	}
+
+	err = s.ruleStore.WriteRuleSet(mutable)
+	if err != nil {
+		return nil, s.handleUpstreamError(err)
+	}
+
+	crs, err := s.FetchRuleSet(namespaceID)
```
xichen2020: The reloading logic below is repeated in `CreateMappingRule`. Can we move it to a separate function `reloadMappingRule()`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
```
xichen2020: The reloading logic below is repeated in `CreateRollupRule`. Can we move it to a separate function `reloadRollupRule()`?

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+	hist, exists := rrs[rollupRuleID]
+	if !exists {
+		return nil, s.rollupRuleNotFoundError(namespaceID, rollupRuleID)
+	}
+	return hist, nil
+}
+
+func (s *store) rollupRuleNotFoundError(namespaceID, ruleID string) error {
+	return r2.NewNotFoundError(
+		fmt.Sprintf("rollup rule: %s doesn't exist in Namespace: %s",
```
xichen2020: nit: take it or leave it, but I usually keep the args on separate lines from the parens.

r2/kv/store.go
```diff
@@ -21,76 +21,382 @@
...
+func (s *store) mappingRuleNotFoundError(namespaceID, ruleID string) error {
+	return r2.NewNotFoundError(
+		fmt.Sprintf("mapping rule: %s doesn't exist in Namespace: %s",
+			ruleID,
+			namespaceID,
+		),
+	)
+}
+
+func (s *store) handleUpstreamError(err error) error {
```
xichen2020, line 389: Do a nil check here just to be safe?

r2/service.go
```diff
@@ -175,24 +174,20 @@ func NewService(
 	return &service{iOpts: iOpts, store: store, authService: authService, rootPrefix: rootPrefix}
 }
 
-func (s *service) write(w http.ResponseWriter, data interface{}) error {
+func (s *service) write(w http.ResponseWriter, retCode int, data interface{}) error {
```
xichen2020: nit: `statusCode` is probably more descriptive.

r2/service.go
```diff
@@ -175,24 +174,20 @@ func NewService(
...
 	if j, err := json.Marshal(data); err == nil {
-		return write(w, j, http.StatusOK)
+		return write(w, j, retCode)
 	}
 	return writeAPIResponse(w, http.StatusInternalServerError, "Could not create response object")
```
xichen2020: nit: `Could` -> `could`

r2/service.go
xichen2020: nit: `write` -> `sendResponse`?

r2/service.go
```diff
@@ -509,38 +549,62 @@ type apiResponse struct {
...
-		ID:            mrv.ID,
-		Name:          mrv.Name,
-		Filters:       mrv.Filters,
-		Policies:      mrv.Policies,
-		CutoverMillis: xtime.ToUnixMillis(xtime.FromNanoseconds(mrv.CutoverNanos)),
+		ID:                  mrv.ID,
+		Name:                mrv.Name,
+		Filters:             mrv.Filters,
+		Policies:            mrv.Policies,
+		CutoverMillis:       xtime.ToUnixMillis(xtime.FromNanoseconds(mrv.CutoverNanos)),
```
xichen2020: nit: can just divide by `time.Millisecond / time.Nanosecond` to be more efficient.

r2/service.go
```diff
@@ -592,25 +658,28 @@ func newRollupRuleJSON(rrv *rules.RollupRuleView) rollupRuleJSON {
...
-		ID:            rrv.ID,
-		Name:          rrv.Name,
-		Filters:       rrv.Filters,
-		Targets:       targets,
-		CutoverMillis: xtime.ToUnixMillis(xtime.FromNanoseconds(rrv.CutoverNanos)),
+		ID:                  rrv.ID,
+		Name:                rrv.Name,
+		Filters:             rrv.Filters,
+		Targets:             targets,
+		CutoverMillis:       xtime.ToUnixMillis(xtime.FromNanoseconds(rrv.CutoverNanos)),
```
xichen2020: Same nit as above.

r2/store.go
```diff
@@ -25,11 +25,11 @@ import "github.com/m3db/m3metrics/rules"
 // Store is a construct that can perform operations against a backing rule store.
 type Store interface {
 	// Fetch namespaces and their version
-	FetchNamespaces() ([]string, int, error)
+	FetchNamespaces() (*rules.NamespacesView, error)
 
-	CreateNamespace(namespaceID string) (string, error)
+	CreateNamespace(namespaceID string, author string) (*rules.NamespaceView, error)
```
xichen2020: Might be good to take an options struct here in case we need more information to be added so it doesn't break the API. Same applies to the rest of the places where we pass in the `author` explicitly.

services/r2ctl/config/r2ctl.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/m3db/m3ctl/r2"
+	"github.com/m3db/m3ctl/r2/kv"
+	"github.com/m3db/m3ctl/r2/stub"
+	"github.com/m3db/m3metrics/rules"
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
+)
+
+// R2CtlConfiguration is the global configuration for r2ctl.
+type R2CtlConfiguration struct {
```
xichen2020: nit: Just call this Configuration, otherwise you end up with `r2ctl.R2CtlConfiguration` which is verbose.

services/r2ctl/config/r2ctl.go
```diff
@@ -0,0 +1,108 @@
...
+	// Simple Auth Config.
+	Auth *auth.SimpleAuthConfig `yaml:"auth"`
+}
+
+// R2StoreConfiguration has all the fields necessary for an R2Store
+type R2StoreConfiguration struct {
+	// Stub means use the stub store.
+	Stub bool `yaml:"stub"`
+
+	// KVStoreConfig is the configuration for the etcd backed implementation of the kvStore.
```
xichen2020: `KV` is the configuration ...

services/r2ctl/config/r2ctl.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// NewR2Store creates
+func (c R2StoreConfiguration) NewR2Store(iOpts instrument.Options) (r2.Store, error) {
+	if c.Stub {
+		return stub.NewStore(iOpts), nil
+	}
+
+	if c.KV == nil {
+		return nil, errors.New("must provide kv configuration if not using stub store")
```
xichen2020: nit: make the error a variable

services/r2ctl/main/main.go
```diff
@@ -118,22 +76,39 @@ func main() {
...
+	if cfg.Auth != nil {
+		authService = cfg.Auth.NewSimpleAuth()
+	}
 
-	doneCh := make(chan struct{})
 	r2ApiServer := server.NewServer(
 		listenAddr,
 		serverOpts,
-		r2.NewService(instrumentOpts, "/r2/v1/", authService, stub.NewStore(instrumentOpts)),
+		r2.NewService(instrumentOpts, "/r2/v1/", authService, store),
```
xichen2020: nit: make `/r2/v1/` a const.

r2/kv/store.go
dgromov: sure

dgromov: cool

r2/kv/store.go
dgromov: yeah

r2/kv/store.go
dgromov: sure

r2/kv/store.go
dgromov: The add will fail if the namespace doesn't exist.

r2/kv/store.go
dgromov: Because I sometimes forgot to press cmd-v

r2/kv/store.go
dgromov: The api we designed calls for newly created or modified objects to be returned on a successful write. The UI will always have to do this so it makes sense to skip the extra work of re-requesting everything.

r2/kv/store.go
dgromov: Will bring this in once the type in m3metrics is correct.

r2/kv/store.go
dgromov: This latest snapshot of the namespace.

r2/service.go
dgromov: cool

glide.lock
dgromov: Spoke offline. This is fine.

r2/kv/store.go
xichen2020: Hm I don't follow, does `ReadNamespaces` return an error in that case? If it does, seems the `CreateNamespace` method would return an error when the namespace doesn't exist, which is probably not what we want?

r2/kv/store.go
xichen2020: I'd think if UI auto refreshes after success this would make the backend code less cluttered but ok.

r2/kv/store.go
xichen2020: Okay cool, can you add a comment here explaining what the `0` is?

r2/kv/store.go
```diff
@@ -21,76 +21,374 @@
...
+	iOpts        instrument.Options
+	updateHelper rules.RuleSetUpdateHelper
+	ruleStore    rules.Store
+	clockOpts    clock.Options
 }
 
 // NewStore returns a new service that knows how to talk to a kv backed r2 store.
-func NewStore(cfg StoreConfig) r2.Store {
-	return &store{cfg: cfg}
+func NewStore(iOpts instrument.Options, uh rules.RuleSetUpdateHelper, rs rules.Store, cOpts clock.Options) r2.Store {
```
xichen2020: nit: might wanna have a `storeOptions` struct containing instrument options and clock options to avoid passing in a ton of args.

r2/kv/store.go
```diff
@@ -21,76 +21,374 @@
...
+	return &store{iOpts: iOpts, updateHelper: uh, ruleStore: rs, clockOpts: cOpts}
 }
 
-func (s *store) FetchNamespaces() ([]string, int, error) {
-	return nil, 0, nil
+func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
+	nss, err := s.ruleStore.ReadNamespaces()
+	if err != nil {
+		return nil, err
```
xichen2020: Should this also be `return nil, s.handleUpstreamError(err)`?

r2/kv/store.go
```diff
@@ -21,76 +21,374 @@
...
+	return &rules.NamespacesView{
+		Namespaces: liveNss,
+		Version:    namespaces.Version,
+	}, nil
 }
 
-func (s *store) CreateNamespace(namespaceID string) (string, error) {
-	return "", nil
+func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*rules.NamespaceView, error) {
+	nss, err := s.ruleStore.ReadNamespaces()
```
xichen2020, line 74: Does `ReadNamespaces` return an error in that case? If it does, seems the `CreateNamespace` method would return an error when the namespace doesn't exist, which is probably not what we want?

r2/service.go
```diff
@@ -536,11 +611,13 @@ func (m mappingRuleJSON) mappingRuleView() *rules.MappingRuleView {
...
-		ID:            mrv.ID,
-		Name:          mrv.Name,
-		Filters:       mrv.Filters,
-		Policies:      mrv.Policies,
-		CutoverMillis: xtime.ToUnixMillis(xtime.FromNanoseconds(mrv.CutoverNanos)),
+		ID:                  mrv.ID,
+		Name:                mrv.Name,
+		Filters:             mrv.Filters,
+		Policies:            mrv.Policies,
+		CutoverMillis:       mrv.CutoverNanos / int64(time.Millisecond),
```
xichen2020: nit: perhaps slightly more readable to do
```
nanosPerMilli = int64(time.Millisecond / time.Nanosecond)
```
and then here do
`CutoverMillis: mrs.CutoverNanos / nanosPerMilli,`

r2/service.go
```diff
@@ -536,11 +611,13 @@ func (m mappingRuleJSON) mappingRuleView() *rules.MappingRuleView {
...
+		LastUpdatedBy:       mrv.LastUpdatedBy,
+		LastUpdatedAtMillis: mrv.LastUpdatedAtNanos / int64(time.Millisecond),
```
xichen2020: Same as above

r2/service.go
```diff
@@ -592,11 +671,13 @@ func newRollupRuleJSON(rrv *rules.RollupRuleView) rollupRuleJSON {
...
-		ID:            rrv.ID,
-		Name:          rrv.Name,
-		Filters:       rrv.Filters,
-		Targets:       targets,
-		CutoverMillis: xtime.ToUnixMillis(xtime.FromNanoseconds(rrv.CutoverNanos)),
+		ID:                  rrv.ID,
+		Name:                rrv.Name,
+		Filters:             rrv.Filters,
+		Targets:             targets,
+		CutoverMillis:       rrv.CutoverNanos / int64(time.Millisecond),
```
xichen2020: Same as above

r2/service.go
```diff
@@ -592,11 +671,13 @@ func newRollupRuleJSON(rrv *rules.RollupRuleView) rollupRuleJSON {
...
+		LastUpdatedBy:       rrv.LastUpdatedBy,
+		LastUpdatedAtMillis: rrv.LastUpdatedAtNanos / int64(time.Millisecond),
```
xichen2020: Same as above

r2/update_options.go
```diff
@@ -0,0 +1,49 @@
...
+// THE SOFTWARE.
+
+package r2
+
+// UpdateOptions is a set of ruleset or namespace update options.
+type UpdateOptions interface {
+	// SetAuthor sets the author.
+	SetAuthor(value string) UpdateOptions
+
+	// ReadTimeout returns the read timeout.
```
xichen2020: Comment should be `// Author returns the author`.

services/r2ctl/config/r2ctl.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Simple Auth Config.
+	Auth *auth.SimpleAuthConfig `yaml:"auth"`
+
+	// Clock configuration used by the api.
+	Clock clock.Configuration
+}
+
+// R2StoreConfiguration has all the fields necessary for an R2Store
+type R2StoreConfiguration struct {
```
xichen2020: Do you need to export this?

services/r2ctl/config/r2ctl.go
```diff
@@ -0,0 +1,116 @@
...
+	// Stub means use the stub store.
+	Stub bool `yaml:"stub"`
+
+	// KV is the configuration for the etcd backed implementation of the kvStore.
+	KV *KVStoreConfig `yaml:"kv,omitempty"`
+}
+
+// KVStoreConfig is the configuration for the etcd backed implementation of the kvStore.
+type KVStoreConfig struct {
```
xichen2020, line 74: Do you need to export this?

r2/kv/store.go
dgromov: ReadNamespaces returns all of the available namespaces, it will not error if a single namespace doesn't exist. The error case there is that the top level namespaces path doesn't exist.

CreateNamespace will error if it encounters an error that is not Namespace not found.

services/r2ctl/config/r2ctl.go
dgromov: For the use case today no, but I figured it would be nice if someone in the open source world wanted to create their own config. Same answer for the other one.

r2/kv/options.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type storeOptions struct {
+	instrumentOpts instrument.Options
+	clockOpts      clock.Options
+}
+
+// NewStoreOptions creates a new set of store options.
+func NewStoreOptions() StoreOptions {
+	return &storeOptions{}
```
xichen2020: Usually we initialize the options to the default options, e.g.,
```
return &storeOptions{
  instrumentOpts: instrument.NewOptions(),
  clockOpts: clock.NewOptions(),
}
```

r2/kv/store.go
```diff
@@ -25,26 +25,23 @@ import (
...
+	sOpts        StoreOptions
 	updateHelper rules.RuleSetUpdateHelper
 	ruleStore    rules.Store
-	clockOpts    clock.Options
 }
 
 // NewStore returns a new service that knows how to talk to a kv backed r2 store.
-func NewStore(iOpts instrument.Options, uh rules.RuleSetUpdateHelper, rs rules.Store, cOpts clock.Options) r2.Store {
-	return &store{iOpts: iOpts, updateHelper: uh, ruleStore: rs, clockOpts: cOpts}
+func NewStore(sOpts StoreOptions, uh rules.RuleSetUpdateHelper, rs rules.Store) r2.Store {
```
xichen2020: The convention is to pass in the options as the last parameter.

r2/kv/store.go
```diff
@@ -359,7 +356,7 @@ func (s *store) FetchRollupRuleHistory(namespaceID, rollupRuleID string) ([]*rul
 }
 
 func (s *store) newUpdateMeta(uOpts r2.UpdateOptions) rules.UpdateMetadata {
-	return s.updateHelper.NewUpdateMetadata(s.clockOpts.NowFn()().UnixNano(), uOpts.Author())
+	return s.updateHelper.NewUpdateMetadata(s.sOpts.ClockOptions().NowFn()().UnixNano(), uOpts.Author())
```
xichen2020: nit: Conventionally we store the clock options as a cached field in the store so you don't need to make a function call every time.

r2/service.go
```diff
@@ -66,6 +65,8 @@ var (
 	rollupRuleRoot        = fmt.Sprintf("%s/%s", namespacePrefix, rollupRulePrefix)
 	rollupRuleWithIDPath  = fmt.Sprintf("%s/{%s}", rollupRuleRoot, ruleIDVar)
 	rollupRuleHistoryPath = fmt.Sprintf("%s/history", rollupRuleWithIDPath)
+
+	nanosPerMilli = int64(time.Millisecond / time.Nanosecond)
```
xichen2020: nit: this should be a const.

r2/kv/store.go
```diff
@@ -109,7 +109,8 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*ru
 		return nil, r2.NewNotFoundError(fmt.Sprintf("namespace: %s does not exist", namespaceID))
 	}
 
-	view, err := ns.NamespaceView(0)
+	// Get the latest view of the namespace.
+	view, err := ns.NamespaceView(len(ns.Snapshots()) - 1)
```
xichen2020: I think we might need to clean up the views at some point, right now some views are sorted by time in descending order, others are sorted by time in ascending order, which is confusing.

xichen2020 (approved):
LGTM

PR #15, opened 2017-10-12
Add validation configuration

xichen2020:
cc @dgromov @cw9 @jeromefroe 

This PR pulls in the latest m3metrics and sets up the validation configuration to perform validation on rule changes.

### DIFF
--- a/services/r2ctl/config/r2ctl.go
+++ b/services/r2ctl/config/r2ctl.go
@@ -30,8 +30,10 @@ import (
 	"github.com/m3db/m3ctl/r2"
 	"github.com/m3db/m3ctl/r2/kv"
 	"github.com/m3db/m3ctl/r2/stub"
+	"github.com/m3db/m3metrics/filters"
+	"github.com/m3db/m3metrics/metric"
+	"github.com/m3db/m3metrics/policy"
 	"github.com/m3db/m3metrics/rules"
-	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3x/log"
 )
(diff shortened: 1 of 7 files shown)


services/r2ctl/config/r2ctl.go
```diff
@@ -52,31 +54,41 @@ type Configuration struct {
...
 	Auth *auth.SimpleAuthConfig `yaml:"auth"`
-
-	// Clock configuration used by the api.
-	Clock clock.Configuration `yaml:"clock"`
 }
 
-// R2StoreConfiguration has all the fields necessary for an R2Store
-type R2StoreConfiguration struct {
+// r2StoreConfiguration has all the fields necessary for an R2 store.
+type r2StoreConfiguration struct {
```
dgromov, line 64: so if anyone wants to run a different main, they have to completely redefine all of these configs?

services/r2ctl/main/main.go
```diff
@@ -90,12 +89,10 @@ func main() {
 		SetMetricsSamplingRate(cfg.Metrics.SampleRate()).
 		SetReportInterval(cfg.Metrics.ReportInterval())
 
-	clockOpts := clock.NewOptions()
-
 	listenAddr := fmt.Sprintf("%s:%d", cfg.HTTP.Host, cfg.HTTP.Port)
 	serverOpts := cfg.HTTP.NewHTTPServerOptions(instrumentOpts)
 
-	store, err := cfg.Store.NewR2Store(instrumentOpts, clockOpts)
```
dgromov, line 98: what happened to the clockOpts?

services/r2ctl/config/r2ctl.go
```diff
@@ -52,31 +54,41 @@ type Configuration struct {
...
-	// KV namespace.
-	RulesKVConfig clusterKV.Configuration `yaml:"rulesKVConfig"`
+// kvStoreConfig is the configuration for the kv backed implementation of the R2 store.
+type kvStoreConfig struct {
+	// KVClient configures the client for key value store.
+	KVClient *etcd.Configuration `yaml:"kvClient" validate:"nonzero"`
 
-	// KVClient configuration for key value store.
-	KVClient *etcd.Configuration `yaml:"etcd" validate:"nonzero"`
+	// KV configures the namespace and the environment.
```
dgromov: s/KV/KVConfig/

services/r2ctl/config/r2ctl.go
xichen2020: The idea is that the users wouldn't need to run their own main file if the config here is flexible enough, and I'd rather keep it that way.

services/r2ctl/config/r2ctl.go
xichen2020: Sure sounds good.

services/r2ctl/main/main.go
xichen2020: It's initialized to `clock.NewOptions()` by default and it's primarily useful to testing. I don't imagine we'll need to configure it in the config, and if we do, we can always add them back.

dgromov (approved):

PR #18, opened 2017-10-13
UI Fixes and Updates

saminzadeh:
The PR covers the following:

- Enhancements to the Policy Editor
  - Agg function selector
  - Better validation for metric type
- Updates to text labels (eg. Name -> Rule Name, Tags -> Rollup Tags)
- Added "Last Updated By" field 
- [Bug] Fixes modal closing bug
- [Bug] Fix history table to display all results
- Move off Antd forms to formik

Policy Editor GIF:
oct-13-2017 11-16-40


ui/src/components/PolicyEditor.js
```diff
@@ -0,0 +1,226 @@
...
+    <div
+      className="clearfix p1"
+      style={{
+        border: '1px solid whitesmoke',
+        lineHeight: 'normal',
+      }}>
+      {_.map(policies, (p, i) => {
+        return (
+          <div
+            key={i}
```
drewproud: Minor nit: might want to make the `key={p}` here instead of the index.

ui/src/components/RollupRuleEditor.js
```diff
@@ -22,67 +22,78 @@ import React from 'react';
...
+            required
+            colon={false}
+            label="Metric Filter"
+            help="eg. tag1:value1 tag2:value2"
+            {...formItemLayout}>
+            <Input
+              name="filter"
+              autoComplete="off"
+              value={values.filter}
+              onChange={(...e) => {
```
drewproud: Is there a better place to drop this logic? Looks difficult to test inline here. If we want to leave it here, it might be worth throwing an Enzyme test around.

drewproud (approved):
Looks great, left some non-blocking feedback.

ui/src/components/PolicyEditor.js
saminzadeh: Ok so it looks like React won't complain if you pass an object, but the [docs](https://reactjs.org/docs/lists-and-keys.html) mention to use a string. I can do a stringify on the object, since the equality check may fail even if the objects are technically the same from a new reference.

saminzadeh:
Closing PR since new commits are getting updated, will start a new one

PR #19, opened 2017-10-13
Pull in latest m3metrics dependency for better policy string representation

xichen2020:
cc @cw9 @dgromov 

This PR pulls in latest m3metrics dependency for better policy string representation.


cw9 (approved):
LGTM

PR #24, opened 2017-10-17
Fix labels, modal refresh, policy editor default tags

saminzadeh:


ui/src/components/RollupRuleEditor.js
```diff
@@ -123,21 +134,27 @@ const TargetsEditorBase = props => {
                 <Icon type="delete" />
               </a>
             }>
-            <label>Metric Name</label>
+            <label>New Metric Name</label>
```
xichen2020: Do you want to call this `New Rollup Metric Name`?

ui/src/components/RollupRulesTable.js
```diff
@@ -38,7 +38,7 @@ function TargetPreview({target}) {
         <div>{target.policies.join(' ')}</div>
       </div>
       <div className="pb1">
-        <label className="bold pb1">Tags</label>
+        <label className="bold pb1">GroupBy Tags</label>
```
xichen2020, line 41: Let's still call this `Rollup Tags` for now and see whether the user finds it confusing.

saminzadeh: sg

ui/src/components/RollupRuleEditor.js
saminzadeh: Sure, that works. Might be worth to user test these types of things. It's hard to say with our tunnel vision now lol

xichen2020: Agreed =p

xichen2020 (approved):
LGTM

PR #26, opened 2017-10-18
Use code instead of pre for table formatting

saminzadeh:
This should fix the table formatting issue to wrap the "Metric Filter" column in the table.


xichen2020 (approved):
LGTM

PR #27, opened 2017-10-18
Adds swagger description fields to MappingRule and RollupRule

m-sandusky:
Adds swagger description fields to MappingRule and RollupRule and their corresponding data models (MappingRuleData, RollupRuleData)

### DIFF
--- a/public/r2/v1/swagger/swagger.json
+++ b/public/r2/v1/swagger/swagger.json
@@ -836,6 +836,9 @@
                 "name": {
                     "type": "string"
                 },
+                "description": {
+                    "type": "string"
+                },
                 "filter": {
                     "type": "string"
                 },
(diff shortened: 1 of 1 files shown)


public/r2/v1/swagger/swagger.json
```diff
@@ -836,6 +836,9 @@
                 "name": {
                     "type": "string"
                 },
+                "description": {
```
xichen2020, line 839: For this we actually need to make a change to the backend protobuf message.

m-sandusky: Yup, I wanted to make a separate pull request for that (as it's in a different repo I believe). This pull request was mainly to confirm  agree on the upcoming API change, and have the changes updated in Swagger.

saminzadeh: Cool, looks good. I agree with Xi, we should merge it with the actual implementation so its consistent on master.

Also, I wonder if there is a way to extend an Object in swagger markup. eg MappingRule is implemented by MappingRuleData. Might help on code duplication.

PR #28, opened 2017-10-24
Update to include latest version of m3metrics

m-sandusky:


m-sandusky:
Just fyi, not sure if runeFrom is in the right file but I didn't see a util file in the repo @xichen2020

services/r2ctl/config/r2ctl.go
```diff
@@ -195,3 +199,11 @@ type policiesConfiguration struct {
 	StoragePolicies  []policy.StoragePolicy   `yaml:"storagePolicies"`
 	AggregationTypes []policy.AggregationType `yaml:"aggregationTypes"`
 }
+
+func runesFrom(s string) []rune {
```
xichen2020: Perhaps `DecodeRuneInString` is more portable?

services/r2ctl/config/r2ctl.go
```diff
@@ -195,3 +199,11 @@ type policiesConfiguration struct {
 	StoragePolicies  []policy.StoragePolicy   `yaml:"storagePolicies"`
 	AggregationTypes []policy.AggregationType `yaml:"aggregationTypes"`
 }
+
+func decodeRuneInString(s string) []rune {
```
xichen2020: nit: `toRunes` might be a better name. Sorry about going back and forth on this.

services/r2ctl/config/r2ctl.go
```diff
@@ -195,3 +199,11 @@ type policiesConfiguration struct {
...
+	var r []rune
```
xichen2020: nit: preallocate enough so we don't need to reallocate, i.e., `r := make([]rune, 0, len(s))`. It might over-allocate a little in case of multi-byte runes but should work in most cases.

services/r2ctl/config/r2ctl.go
m-sandusky: No worries, makes sense.

services/r2ctl/config/r2ctl.go
m-sandusky: Good point, will do

xichen2020 (approved):
LGTM

PR #31, opened 2017-10-27
Adds instrumentation for reporting service metrics.

m-sandusky:
This is part 1 of a 2 part change. This adds the instrumentation to metrics logging in the service API. The second part will wrap the route logic for each endpoint with this logic in order to actually report the metrics.


m-sandusky:
Build failures due to failing lint checks (unused functions and structs). Want to get feedback for part 1. Once addressed, I'll add part 2 to this pull request which should address these build failures.

r2/service.go
```diff
@@ -28,8 +28,11 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/uber-go/tally"
```
xichen2020: nit: convention is to include 3rd party imports at the bottom part of the import block.

r2/service.go
```diff
@@ -173,7 +233,13 @@ func NewService(
 	authService auth.HTTPAuthService,
 	store Store,
 ) server.Service {
-	return &service{iOpts: iOpts, store: store, authService: authService, rootPrefix: rootPrefix}
+	return &service{
+		iOpts:          iOpts,
+		store:          store,
+		authService:    authService,
+		rootPrefix:     rootPrefix,
+		serviceMetrics: serviceMetrics{scope: iOpts.MetricsScope()},
```
xichen2020: nit: take it or leave it, but the convention is to have a ctor that returns a serviceMetrics struct.

r2/service.go
```diff
@@ -160,10 +219,11 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
...
 type service struct {
-	iOpts       instrument.Options
-	rootPrefix  string
-	store       Store
-	authService auth.HTTPAuthService
+	iOpts          instrument.Options
+	rootPrefix     string
+	store          Store
+	authService    auth.HTTPAuthService
+	serviceMetrics serviceMetrics
```
xichen2020: nit: `serviceMetrics` -> `metrics`.

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+	}
+	return data, nil
+}
+
+func (sm *serviceMetrics) recordMetric(path, method, namespace string, d time.Duration, err error) {
+	status := "success"
+	if err != nil {
+		status = "error"
+	}
+	endpoint := path + method + status
```
xichen2020: Perhaps it's better to create a struct like so:
``` 
type serviceMetricID struct {
  Path string
  Method string
  Status string
}
``` 
and use that as map key instead of the concatenated string?

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+
+	sm.Lock()
```
xichen2020: Might be more efficient to do double locking here like below
```
sm.RLock()
m, exists := sm.metrics[endpoint]
sm.RUnlock()
if exists {
  m.Record(...)
  return
}

sm.Lock()
defer sm.Unlock()
m, exists = sm.metrics[endpoint]
if exists {
  m.Record(...)
  return
}
m = ...
sm.metrics[endpoint] = m
m.Record(...)
```

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+
+type serviceMetrics struct {
+	sync.RWMutex
+	scope tally.Scope
+
+	// Endpoints mapped to metrics (endpoint = path + method + status).
+	metrics map[string]serviceMetric
+}
+
+type serviceMetric struct {
```
xichen2020: Perhaps define a method `Record` here:
```
func (m *serviceMetric) Record(d time.Duration) {
  m.timer.Record(d)
  m.counter.Inc(1)
}
```
so you can simply call `m.Record(d)` in `handleRoute`?

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+	timer   tally.Timer
+	counter tally.Counter
+}
+
+// TODO(sandusky) Wrap each route with this function.
+func handleRoute(rf routeFunc, s *service, r *http.Request, namespace string) (interface{}, error) {
+	start := time.Now()
```
xichen2020: I'd pass in a clock.Options (see `m3db/m3x/clock`) to the service constructor, store the `NowFn` in the service, and use the `NowFn` here so you can easily provide a different `nowFn` if needed (e.g., in tests).

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
 	auth  auth.HTTPAuthService
 }
 
+type routeFunc func(s *service, r *http.Request) (data interface{}, err error)
```
xichen2020: Do the handlers for all routes have this signature?

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+	dur := time.Since(start)
+	s.serviceMetrics.recordMetric(r.RequestURI, r.Method, namespace, dur, err)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+func (sm *serviceMetrics) recordMetric(path, method, namespace string, d time.Duration, err error) {
+	status := "success"
```
xichen2020: nit: would make "success" and "error" consts.

r2/service.go
```diff
@@ -122,6 +125,62 @@ type r2Handler struct {
...
+
+	scope := sm.scope
+	m, exists := sm.metrics[endpoint]
+	if !exists {
+		scope = scope.Tagged(map[string]string{
+			"namespace":   namespace,
+			"http-method": method,
+			"route-path":  path,
+		})
+		m = serviceMetric{
```
xichen2020: nit: make this a ctor call.

r2/service.go
m-sandusky: Fixed, I'm using vscode - strange that it put the import by itself up there. I'll if there's a config file in the preferences.

r2/service.go
m-sandusky: I believe they do - the issue here though is that if we want to add 'namespace' to the service metrics, we'll have to read the request twice in multiple places ie 
`
vars := mux.Vars(request)
ns := vars[namespace] 
`

Still trying to think out a good solution to avoiding this.

r2/service.go
m-sandusky: sgtm

r2/service.go
m-sandusky: done

r2/service.go
m-sandusky: Yeah really good point - done.

m-sandusky (commented):
Thank for the review! I've addressed your comments and pulled the metrics related code to it's own file. Once things look good enough - I'll add some tests for these and refactor the routes code so the metrics logic can wrap the existing logic.

r2/metrics.go
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
xichen2020: nit: we tend to keep the metrics structs in the same file as the one the metrics struct is defined for. If not, perhaps rename this file to `r2/service_metrics.go`?

r2/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/uber-go/tally"
+)
+
+const (
+	metricSuccess = "success"
+	metricError   = "error"
+)
+
+type serviceMetrics struct {
+	sync.RWMutex
```
xichen2020: nit: leave a blank line after embedded struct.

r2/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		"http-method": method,
+		"route-path":  path,
+	})
+	m = newServiceMetric(scope, status)
+	sm.metrics[id] = m
+	m.Record(d)
+}
+
+func newServiceMetrics(scope tally.Scope) serviceMetrics {
+	return serviceMetrics{scope: scope}
```
xichen2020: Probably need to initialize the map here? otherwise adding metrics to the map will panic.

r2/service.go
```diff
@@ -172,8 +189,16 @@ func NewService(
 	rootPrefix string,
 	authService auth.HTTPAuthService,
 	store Store,
+	clock clock.Options,
```
xichen2020: nit: conventionally this is called `clockOpts`.

services/r2ctl/main/main.go
```diff
@@ -105,7 +107,13 @@ func main() {
 	r2ApiServer := server.NewServer(
 		listenAddr,
 		serverOpts,
-		r2.NewService(instrumentOpts, r2apiPrefix, authService, store),
+		r2.NewService(
+			instrumentOpts,
+			r2apiPrefix,
+			authService,
+			store,
+			clock.NewOptions().SetNowFn(time.Now),
```
xichen2020: nit: by default the now fn in a clock options is `time.Now` so you don't need to set it explicitly.

r2/service_metrics.go
```diff
@@ -78,7 +79,7 @@ func (sm *serviceMetrics) recordMetric(path, method, namespace string, d time.Du
 }
 
 func newServiceMetrics(scope tally.Scope) serviceMetrics {
-	return serviceMetrics{scope: scope}
+	return serviceMetrics{scope: scope, metrics: map[serviceMetricID]serviceMetric{}}
```
xichen2020: nit: `metrics: make(map[serviceMetricID]serviceMetric)` is a bit more idiomatic in Go.

r2/service_routes_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+func TestHandleRoute(t *testing.T) {
+	s := newTestService()
+	r := newTestGetRequest()
+	expected := newNamespacesJSON(&rules.NamespacesView{})
+	actual, err := handleRoute(fetchNamespaces, s, r, "ns")
+	require.NoError(t, err)
+	require.Equal(t, expected, actual)
+}
+func TestFetchNamespaces_Success(t *testing.T) {
```
xichen2020: We normally don't use underscores in function names.

xichen2020 (approved):
LGTM

PR #32, opened 2017-11-09
Fixes passing nil request into handler

m-sandusky:
Fixes issue where nil was passed instead of the request (probably should never do that :) )


r2/service.go
```diff
@@ -194,8 +194,8 @@ func (s *service) sendResponse(w http.ResponseWriter, statusCode int, data inter
 	return writeAPIResponse(w, http.StatusInternalServerError, "could not create response object")
 }
 
-func (s *service) fetchNamespaces(w http.ResponseWriter, _ *http.Request) error {
-	data, err := handleRoute(fetchNamespaces, s, nil, "")
+func (s *service) fetchNamespaces(w http.ResponseWriter, r *http.Request) error {
```
xichen2020, line 215: Mind adding a regression test for it as well?

xichen2020 (approved):
LGTM

PR #34, opened 2017-11-18
Fix yaml tag for write whitelisted user ids

xichen2020:
cc @m-sandusky @cw9 

This PR fixes the yaml tag for write whitelisted user ids and adds a unit test.


cw9 (approved):
lgtm

PR #36, opened 2017-11-21
Pass error handler into auth handler

jskelcy:


m-sandusky (approved):
Lgtm! I'd possibly add a comment in auth.go quickly discussing the errorResponseHandler but other than that looks good.

PR #37, opened 2017-11-28
Pull in m3metrics to support negation of filter patterns

xichen2020:
cc @cw9 @jskelcy 

This PR pulls in m3metrics to support negation of filter patterns.


cw9 (approved):
lgtm

PR #38, opened 2017-11-28
clean up health check tests

jskelcy:


jskelcy:
I may have changed the health check endpoint on this one. going to look into why the tests are still passing

health/service.go
```diff
@@ -95,7 +90,7 @@ func m3ctlHealthCheck(w http.ResponseWriter, r *http.Request) {
 // RegisterHandlers registers health handlers.
 func (s *service) RegisterHandlers(router *mux.Router) {
 	log := s.iOpts.Logger()
-	router.HandleFunc("", m3ctlHealthCheck)
+	router.HandleFunc(s.URLPrefix(), m3ctlHealthCheck)
```
jskelcy: I am a little confused as to why the health check was at the root, and how this worked, to begin with.

jskelcy:
Fixed the health check issue. I didnt realize it was using a prefixed router.

health/service_test.go
```diff
@@ -42,18 +45,18 @@ func TestHealthCheck(t *testing.T) {
 	rr := httptest.NewRecorder()
 	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
 	// pass 'nil' as the third parameter.
-	req := http.Request{Method: "GET", RequestURI: "/health"}
+	req, err := http.NewRequest("GET", "/health", bytes.NewBuffer(nil))
```
robskillington: You can just pass nil as third arg here, no need to create a new buffer that never get's used.

health/service.go
```diff
@@ -72,15 +72,10 @@ func hostName() string {
 func m3ctlHealthCheck(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
 	host := hostName()
-	h := healthCheckResult{Host: host, Timestamp: start}
 	status := healthCheck()
+	h := healthCheckResult{Host: host, Timestamp: start, Status: status}
 	h.ResponseTime = time.Since(start)
 
-	h.Status = status
```
jskelcy, line 79: There is a slight change to the functionality here where if there is a !ok status we return a 200 with the status as part of the body. Personally, I think this is helpful because you can convey information like being in a bootstrapping state, where the server is up but not ready to receive requests. It is also the case that we hardcode the status to be `ok` which means that under no circumstances can we return a 500 here anyway.

health/service_test.go
jskelcy: Nice. I figured that would error down the line somewhere but it looks like that totally works.

m-sandusky (approved):
lgtm

PR #40, opened 2017-12-13
Update m3cluster and m3metrics to tip of master

xichen2020:
cc @cw9 @m-sandusky 

Updating m3cluster and m3metrics dependencies to tip of master.


cw9 (approved):
LGTM

PR #42, opened 2017-12-19
Distinguish action takers from action originators

xichen2020:
cc @cw9 @m-sandusky @jskelcy 

This PR distinguishes users taking the action from those originating the action, which comes handy when users committing changes are not necessarily who initiated the changes.


auth/simple_test.go
```diff
@@ -150,6 +166,26 @@ func TestAuthenticateFailure(t *testing.T) {
 	require.Equal(t, "application/json", recorder.HeaderMap["Content-Type"][0])
 }
 
+func TestAuthenticateWithOriginatorID(t *testing.T) {
```
cw9, line 169: Also add a test for the case where there is no originator provided?

xichen2020: Yeah that's covered in the testcase above `TestAuthenticationFailure`.

m-sandusky (approved):

PR #46, opened 2017-12-30
Add namespace validation logic

xichen2020:
cc @jeromefroe @m-sandusky  @cw9 

This PR pulls in latest m3metrics change to add namespace validation. It also does some refactoring by moving `Service` interface into its own package with two specific implementations (`r2` and `health`), as well as defining the `Server` interface with a specific HTTP implementation, to make the repo layout easier to understand.


server/http/server.go
```diff
@@ -0,0 +1,113 @@
...
+type server struct {
+	server   *http.Server
+	services []service.Service
+	logger   log.Logger
+	wg       sync.WaitGroup
+}
+
+// NewServer creates a new HTTP server.
+func NewServer(address string, opts Options, services ...service.Service) mserver.Server {
+	cloned := make([]service.Service, len(services))
```
jeromefroe: Why do we need to make a copy here?

server/server.go
```diff
@@ -16,38 +16,17 @@
...
-// THE SOFTWARE.
+// THE SOFTWARE
 
 package server
 
-import (
-	"net/http"
+// Server is a server capable of listening to incoming traffic and closing itself
+// when it's shut down.
+type Server interface {
```
jeromefroe, line 25: Why make this into an interface instead of leaving as a struct? I don't have much context, but it's seems like there would only ever be one server, no? This is the [approach taken by the Go standard library](https://golang.org/pkg/net/http/#Server) as well.

server/http/server.go
xichen2020: This is to prevent the list of services from being mutated externally after the server is constructed, e.g., see https://play.golang.org/p/YIZLwWGkmv0

server/server.go
xichen2020: I could've, but it'll make testing more difficult because you can't mock it as a struct. Go standard library worked around this by providing `httptest`, but that's a completely different approach from what we usually do for unit testing.

Was also thinking maybe at some point there could be a GRPC server for communicating with internal services.

server/http/server.go
jeromefroe: Ha, that is weird. Maybe add a comment just to make it clear?

jeromefroe (approved):
LGTM

server/http/server.go
xichen2020: Yeah this is essentially the compiler trying to avoid extra allocation with variadic functions but yeah could potentially cause surprises, I'll add a comment.

PR #47, opened 2017-12-31
Pull in m3metrics to validate filter early

xichen2020:
cc @jeromefroe @m-sandusky @cw9 

This PR pulls in the m3metrics PR to validate filter during snapshot construction. It also determines whether the error returned upstream is a contained error and extracts the inner error if so to avoid losing the original error type.


jeromefroe (approved):
LGTM

PR #50, opened 2018-01-03
Add validate namespace endpoint to m3ctl

m-sandusky:
This PR adds a namespace validation endpoint to m3ctl. Will probably add a couple more tests before merging.


service/r2/kv/store.go
```diff
@@ -35,17 +35,19 @@ type store struct {
 	opts         StoreOptions
 	ruleStore    rules.Store
 	updateHelper rules.RuleSetUpdateHelper
+	validator    rules.Validator
 }
 
 // NewStore returns a new service that knows how to talk to a kv backed r2 store.
-func NewStore(rs rules.Store, opts StoreOptions) r2.Store {
+func NewStore(rs rules.Store, v rules.Validator, opts StoreOptions) r2.Store {
```
xichen2020: Perhaps make `validator` part of the `StoreOptions`?

service/r2/kv/store.go
```diff
@@ -73,6 +75,10 @@ func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
 	}, nil
 }
 
+func (s *store) ValidateNamespace(rs *rules.RuleSetSnapshot) error {
```
xichen2020: Perhaps call this `ValidateRuleSet` to be more accurate?

service/r2/kv/store.go
```diff
@@ -73,6 +75,10 @@ func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
...
+	return s.validator.ValidateSnapshot(rs)
```
xichen2020: I'd check if `s.validator` is nil and return an `errNilValidator = errors.New("validator not provided")` just to be defensive.

service/r2/routes.go
```diff
@@ -64,6 +64,27 @@ func createNamespace(s *service, r *http.Request) (data interface{}, err error)
...
 
+func validateNamespace(s *service, r *http.Request) (data interface{}, err error) {
+	vars := mux.Vars(r)
+	var rsj ruleSetJSON
+	if err := parseRequest(&rsj, r.Body); err != nil {
+		return nil, err
+	}
+	if vars[namespaceIDVar] != rsj.Namespace {
+		return nil, fmt.Errorf(
+			"NamespaceID param: %s and RuleSet namespaceID: %s do not match",
```
xichen2020: nit: error messages should start with lowercase letters. Also might be easier on the eyes to say
`namespaceID param %s and ruleset namespaceID %s do not match`?

service/r2/routes.go
```diff
@@ -64,6 +64,27 @@ func createNamespace(s *service, r *http.Request) (data interface{}, err error)
...
+			vars[namespaceIDVar],
+			rsj.Namespace,
+		)
+	}
+
+	rss := rsj.ruleSetSnapshot()
+	if err := s.store.ValidateNamespace(rss); err != nil {
+		return nil, err
+	}
+	return fmt.Sprintf("Namespace and rule-set are valid"), nil
```
xichen2020: Perhaps just return `Ruleset is valid`?

service/r2/routes.go
```diff
@@ -64,6 +64,27 @@ func createNamespace(s *service, r *http.Request) (data interface{}, err error)
 	return newNamespaceJSON(view), nil
 }
 
+func validateNamespace(s *service, r *http.Request) (data interface{}, err error) {
```
xichen2020: Perhaps `validateRuleSet` is more accurate?

service/r2/service.go
```diff
@@ -192,6 +193,7 @@ func (s *service) RegisterHandlers(router *mux.Router) {
 
 	router.Handle(namespacePath, h.wrap(s.fetchNamespaces)).Methods(http.MethodGet)
 	router.Handle(namespacePath, h.wrap(s.createNamespace)).Methods(http.MethodPost)
+	router.Handle(namespaceValidatePath, h.wrap(s.validateNamespace)).Methods(http.MethodPost)
```
xichen2020: How is auth going to work for validation endpoint? Is it "if a user has read access, then he/she can validate?" Or does he/she need to have write access to validate?

service/r2/service.go
```diff
@@ -553,3 +564,20 @@ type ruleSetJSON struct {
...
+func (r ruleSetJSON) ruleSetSnapshot() *rules.RuleSetSnapshot {
+	rss := rules.RuleSetSnapshot{
+		Namespace:    r.Namespace,
+		Version:      r.Version,
+		MappingRules: map[string]*rules.MappingRuleView{},
+		RollupRules:  map[string]*rules.RollupRuleView{},
+	}
+
+	for _, mr := range r.MappingRules {
+		rss.MappingRules[mr.Name] = mr.mappingRuleView()
```
xichen2020: Hm we should use the UUID rather than the name no?

service/r2/service.go
```diff
@@ -553,3 +564,20 @@ type ruleSetJSON struct {
...
+	}
+	for _, rr := range r.RollupRules {
+		rss.RollupRules[rr.Name] = rr.rollupRuleView()
```
xichen2020: Hm we should use the UUID rather than the name no?

service/r2/stub/store.go
```diff
@@ -304,6 +304,11 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*ru
 	}
 }
 
+func (s *store) ValidateNamespace(rs *rules.RuleSetSnapshot) error {
+	// Assumes no validation config for stub store so all rule sets are valid
```
xichen2020: nit: add a period at the end.

service/r2/store.go
```diff
@@ -27,6 +27,9 @@ type Store interface {
 	// FetchNamespaces fetches namespaces.
 	FetchNamespaces() (*rules.NamespacesView, error)
 
+	// ValidateNamespace validates a namespace's ruleset.
+	ValidateNamespace(rs *rules.RuleSetSnapshot) error
```
xichen2020: `ValidateRuleSet` is probably more accurate in this  case.

service/r2/routes.go
```diff
@@ -64,6 +64,27 @@ func createNamespace(s *service, r *http.Request) (data interface{}, err error)
...
+	}
+	if vars[namespaceIDVar] != rsj.Namespace {
+		return nil, fmt.Errorf(
+			"NamespaceID param: %s and RuleSet namespaceID: %s do not match",
+			vars[namespaceIDVar],
+			rsj.Namespace,
+		)
+	}
+
+	rss := rsj.ruleSetSnapshot()
```
jskelcy: NIT:
Given that this function is only called in this file, and ruleSetSnapshot would be the only stuct with methods on it. Perhaps make this function take a ruleSetJSON and return a ruleSetSnapshot. The It might also have it as sort of public util function.

services/r2ctl/config/r2ctl.go
```diff
@@ -131,5 +131,5 @@ func (c kvStoreConfig) NewStore(instrumentOpts instrument.Options) (r2.Store, er
 	r2StoreOpts := r2kv.NewStoreOptions().
 		SetInstrumentOptions(instrumentOpts).
 		SetRuleUpdatePropagationDelay(c.PropagationDelay)
-	return r2kv.NewStore(rulesStore, r2StoreOpts), nil
+	return r2kv.NewStore(rulesStore, validator, r2StoreOpts), nil
```
jskelcy: I dont know enough to say either way, but could the validator go on the r2StoreOptions? If we already have a catch-all for parameters to this function perhaps we should just add it there instead.

service/r2/kv/store.go
jskelcy: would it be reasonable to panic on NewStore if no validator is passed in? We have a stub validator so even for tests there should always be one available.

service/r2/kv/store.go
m-sandusky: I thought about this, and to me if we're going to throw an error on the validator no being present (from the below comment) then it would make sense to have the Validator be a required param. 

However,  if passing a Validator should be optional then the endpoint should have a successful validation for every request if no validator was configured for the store. I've added the logic for this but can change it back.

service/r2/kv/store.go
m-sandusky: That's interesting - essentially we want the validator that is used in the instance of kv.TxnStore in the construction of NewStore to be the same. We allow for a nil validator to be passed in there (and bypass any validation if so) so if that's the case I'd assume everything passed to the validation endpoint should succeed.

service/r2/routes.go
m-sandusky: Should we edit the actual endpoint to reflect this as well? ie (/namespaces/{namespace_id}/validate_rule_set)

service/r2/routes.go
m-sandusky: I was following the general convention with converting the JSON structs to View/Snapshots (for better or for worse), open to change if you feel strongly about it :)

service/r2/service.go
m-sandusky: Yup, changed.

services/r2ctl/config/r2ctl.go
m-sandusky: Yup, added there

m-sandusky (commented):
@xichen2020 @jskelcy Thanks for the review! I've addressed your comments.

service/r2/kv/options.go
```diff
@@ -96,3 +104,13 @@ func (o *storeOptions) SetRuleUpdatePropagationDelay(value time.Duration) StoreO
 func (o *storeOptions) RuleUpdatePropagationDelay() time.Duration {
 	return o.ruleUpdatePropagationDelay
 }
+
+func (o *storeOptions) SetValidator(value rules.Validator) StoreOptions {
+	opts := *o
```
jskelcy, line 109: This appears to be the convention but is there a need for this deref @xichen2020 ?

service/r2/routes.go
```diff
@@ -64,6 +64,27 @@ func createNamespace(s *service, r *http.Request) (data interface{}, err error)
...
+			vars[namespaceIDVar],
+			rsj.Namespace,
+		)
+	}
+
+	rss := rsj.ruleSetSnapshot(ruleSetSnapshotOpts{generateMissingID: true})
+	if err := s.store.ValidateRuleSet(rss); err != nil {
+		return nil, err
+	}
+	return fmt.Sprintf("Ruleset is valid"), nil
```
jskelcy: 1) I think you can just return a string, no need to use fmt.Sprintf

2) Given that this just returns a string that tells the caller that seems a little self-explanatory do you think this function could just return an error?

service/r2/service.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/m3db/m3metrics/rules"
 	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/instrument"
+	"github.com/pborman/uuid"
```
jskelcy: Im double checking on this for OSS but I think we may want to use github.com/satori/go.uuid

service/r2/service.go
```diff
@@ -553,3 +565,37 @@ type ruleSetJSON struct {
 	MappingRules  []mappingRuleJSON `json:"mappingRules"`
 	RollupRules   []rollupRuleJSON  `json:"rollupRules"`
 }
+
+// Creates a new RuleSetSnapshot from a rulesetJSON. If the ruleSetJSON has no IDs for any of its
+// mapping rules or rollup rules, it generates missing IDs and sets as a string UUID string so they
+// can be stored in a mapping (id -> rule).
```
jskelcy: If a rule is being added, not updated is that when there would not be an ID for a rule? I might update the comment to explain what the scenario is where this happens.

service/r2/routes.go
m-sandusky: 1) Thanks - nice catch.
2) Following convention on the other endpoints that return string values in the service (ie deleteMappingRule, deleteRollupRule, etc.) I'm open to whatever either way.

service/r2/service.go
m-sandusky: Yeah, I've used this on in the past as well. m3ctl uses "github.com/pborman/uuid" in other places though so I didn't want add a new dependency for a similar use case. @xichen2020 Do you have any thoughts?

service/r2/service.go
m-sandusky: Left a comment on this earlier - not sure why it didn't post. In any case this is a good point. If the requester does not have write access, due to how the current auth is set up, they will not be validated for the endpoint. The way around this is to make the endpoint GET or to change how we do auth.

service/r2/service.go
m-sandusky: Yeah that's exactly it - I'll add a comment for this scenario.

service/r2/service.go
```diff
@@ -553,3 +565,39 @@ type ruleSetJSON struct {
...
+		RollupRules:  map[string]*rules.RollupRuleView{},
+	}
+
+	for _, mr := range r.MappingRules {
+		id := mr.ID
+		if id == "" && opts.generateMissingID {
+			id = uuid.New()
+			mr.ID = id
+		}
+		rss.MappingRules[id] = mr.mappingRuleView()
```
jskelcy: If the mapping id is an empty string and opts.generateMissingID is false this will continuously overwrite `rss.MappingRules[""]`. Im not sure what the best behavior is, maybe just leave it out of the list? But I think just taking whatever the last rule is and adding that to the snapshot while dropping all the other ones is kind of confusing.

jskelcy (changes requested):
address the empty string rule id issue.

service/r2/service.go
m-sandusky: Updated per in person conversation

service/r2/kv/options.go
xichen2020: Yeah so this is to ensure the options are immutable.

service/r2/kv/store.go
```diff
@@ -73,6 +73,15 @@ func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
 	}, nil
 }
 
+func (s *store) ValidateRuleSet(rs *rules.RuleSetSnapshot) error {
+	validator := s.opts.Validator()
+	// No validator is set so by default the ruleSetSnapshot is valid
+	if validator == nil {
+		return nil
```
xichen2020: Perhaps we should return an error here instead? If we accidentally forget setting the validator in the options, and someone calls the validation endpoint, we should return an error saying the validator is not set instead of returning ok?

service/r2/service.go
xichen2020: Fine w/ either, as long as there's only one for the repo.

Also, please add a blank line between m3db imports and 3rd party imports.

service/r2/service.go
xichen2020: I think we can make the `NewAuthHandler` take an additional parameter `authType`:
```
type authorizationType int

const (
  none authorizationType
  readOnly
  writeOnly
  readWrite
)

func (a simpleAuth) NewAuthHandler(
  authType authorizationType,
  next http.Handler,
  errHandler errorResponseHandler,
) http.Handler {
  ...
}

return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
  ...
  switch authType {
    case none:
      ...
    case readOnly:
      ...
    case writeOnly:
      ...
    case readWrite:
      ...
}
```
Then in `service.go`, register the auth type
```
func registerRoute(router *mux.Router, path, method string, handler http.Handler) error {
  authType, exists := authRegistry[endpoint{Path:path, Method: method}]
  if !exists {
    // Should make this a method instead.
    switch method:
    case http.GET:
      authType = readOnly
    case http.POST, ...:
      authType = writeOnly
    default:
      return errUnknownHTTPMethod
  }
  fn := h.wrap(authType, handler)
  router.Handle(path, fn).Methods(method)
  return nil
}

func (h r2Handler) wrap(authType authorizationType, fn r2HandlerFunc) http.Handler {
	f := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if err := fn(w, r); err != nil {
			h.handleError(w, err)
		}
	})
	return h.auth.NewAuthHandler(authType, f, writeAPIResponse)
}

type endpoint struct {
  Path string
  Method string
}

var authRegistry = map[endpoint]authorizationType{
  {Path: namespaceValidePath, Method: http.POST}: readOnly
}
```

service/r2/service.go
xichen2020: nit: Change the comment to `ruleSetSnapshot creates a ...`.

You might also make this behavior controlled by an enum passed in though (e.g., genIDType), e.g., if this is an update request, we probably shouldn't generate the missing IDs since that should be a validation error.

service/r2/service.go
```diff
@@ -553,3 +565,33 @@ type ruleSetJSON struct {
...
 }
+
+// Creates a new RuleSetSnapshot from a rulesetJSON. If the ruleSetJSON has no IDs for any of its
+// mapping rules or rollup rules, it generates missing IDs and sets as a string UUID string so they
+// can be stored in a mapping (id -> rule).
+func (r ruleSetJSON) ruleSetSnapshot() *rules.RuleSetSnapshot {
+	rss := rules.RuleSetSnapshot{
+		Namespace:    r.Namespace,
+		Version:      r.Version,
+		MappingRules: map[string]*rules.MappingRuleView{},
```
xichen2020: nit: it's more efficient to do the following to avoid repeated map reallocations.
```
mappingRules := make(map[string]*rules.MappingRuleView, len(r.MappingRules))
for _, mr := r.MappingRules {
  ...
}

rollupRules := make(map[string]*rules.RollupRuleView, len(r.RollupRules))
for _, rr := range r.RollupRules {
  ..
}
return rules.RuleSetSnapshot{
  Namespace: r.Namespace,
  Version: r.Version,
  MappingRules: mappingRules,
  RollupRules: rollupRules,
}
```

service/r2/service_test.go
```diff
@@ -0,0 +1,481 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package r2
+
+import (
+	"testing"
+
+	"github.com/pborman/uuid"
```
xichen2020: Reorder the imports.

service/r2/service_test.go
```diff
@@ -0,0 +1,481 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
xichen2020, line 1: Brilliant, thanks for the tests! 👍

auth/simple.go
```diff
@@ -92,14 +92,22 @@ type simpleAuthorization struct {
...
-	switch httpMethod {
-	case http.MethodGet:
+func (a simpleAuthorization) authorize(authType AuthorizationType, userID string) error {
+	switch authType {
+	case AuthorizationTypeNone:
+		return nil
+	case AuthorizationTypeReadOnly:
 		return a.authorizeUser(a.readWhitelistEnabled, a.readWhitelistedUserIDs, userID)
-	case http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete:
+	case AuthorizationTypeWriteOnly:
```
jskelcy: What is a write-only user? Who would fall into this bucket?

jskelcy:
Looks like there is a significant overhaul to how auth works. Can you give a quick rundown on why you made the changes?

service/r2/kv/store.go
```diff
@@ -73,6 +74,15 @@ func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
 	}, nil
 }
 
+func (s *store) ValidateRuleSet(rs *rules.RuleSetSnapshot) error {
+	validator := s.opts.Validator()
+	// If no validator is set, then the validation functionality is not applicable
+	if validator == nil {
+		return errors.New("no validator set on StoreOptions so validation is not applicable")
```
jskelcy: You can use fmt.Errorf here dont need to mess with imports.

xichen2020: The general rule of thumb is actually to only use `fmt.Errorf` when you need to format some values as part of the error message. For errors that are just a const string, `errors.New` is preferred.

auth/simple.go
m-sandusky: Write only users would be those that are say whitelisted to make writes but not whitelisted for reads. I can't really think of a current situation where that would be the case but theoretically it could happen (ie a service that writes data potentially sensitive and doesn't have permissions to access any of the read endpoints). I realize that this doesn't really fit our current uses but this is why it was included.

m-sandusky:
@jskelcy In reference to explanation for the auth revamp:

This was done because previously, the Auth handler would authorize users based on HTTP verb sent by the request. Ie In order to be authorized for a POST request, you needed to be a whiltelisted writer and for GET requests, you needed to be a whitelisted reader. The validation endpoint brought up the situation where it is a POST request but you really shouldn't need write whitelist credentials to access it. Xi and I discussed that certain endpoints can have different needs for auth which is where this notion of authorization type comes into play. Each endpoint has an authorization type which is configured when registering the routes for the service (with defaults falling back to GET -> need read access, POST/PUT/DELETE/PATCH need write credentials).

auth/auth.go
```diff
@@ -27,20 +27,34 @@ import (
...
 type errorResponseHandler func(w http.ResponseWriter, code int, msg string) error
 
 const (
 	// UserIDField is a key
 	UserIDField keyType = iota
 )
 
+const (
+	// AuthorizationTypeNone is the no authorizationType case.
+	AuthorizationTypeNone AuthorizationType = iota
```
xichen2020: nit: These are a bit of verbose, perhaps just `None`, `ReadOnly`, `WriteOnly`, and `ReadWrite`?

auth/noop.go
```diff
@@ -37,7 +37,7 @@ func NewNoopAuth() HTTPAuthService {
 	return noopAuth{}
 }
 
-func (a noopAuth) NewAuthHandler(next http.Handler, errHandler errorResponseHandler) http.Handler {
+func (a noopAuth) NewAuthHandler(authType AuthorizationType, next http.Handler, errHandler errorResponseHandler) http.Handler {
```
xichen2020: nit: omit the parameter name `authType` or use `_` since it's not used here.

auth/simple.go
```diff
@@ -92,14 +92,22 @@ type simpleAuthorization struct {
...
 
-func (a simpleAuthorization) authorize(httpMethod, userID string) error {
-	switch httpMethod {
-	case http.MethodGet:
+func (a simpleAuthorization) authorize(authType AuthorizationType, userID string) error {
+	switch authType {
+	case AuthorizationTypeNone:
+		return nil
+	case AuthorizationTypeReadOnly:
 		return a.authorizeUser(a.readWhitelistEnabled, a.readWhitelistedUserIDs, userID)
```
xichen2020: nit: This can be moved to a `a.authorizeUserForRead` method

auth/simple.go
```diff
@@ -92,14 +92,22 @@ type simpleAuthorization struct {
...
-	case http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete:
+	case AuthorizationTypeWriteOnly:
+		return a.authorizeUser(a.writeWhitelistEnabled, a.writeWhitelistedUserIDs, userID)
```
xichen2020: nit: This can be moved to a `a.authorizeUserForWrite` method.

auth/simple.go
```diff
@@ -92,14 +92,22 @@ type simpleAuthorization struct {
...
+	case AuthorizationTypeReadWrite:
+		err := a.authorizeUser(a.readWhitelistEnabled, a.readWhitelistedUserIDs, userID)
```
xichen2020: nit: This can reuse`a.authorizeUserForRead` and `a.authorizeUserForWrite` methods defined above.

auth/simple.go
```diff
@@ -92,14 +92,22 @@ type simpleAuthorization struct {
...
+		if err != nil {
+			return err
+		}
 		return a.authorizeUser(a.writeWhitelistEnabled, a.writeWhitelistedUserIDs, userID)
 	default:
-		return fmt.Errorf("unsupported request method: %s", httpMethod)
+		return fmt.Errorf("unsupported authorizationType %v passed to handler", authType)
```
xichen2020: nit: `authorizationType` -> `authorization type`

service/r2/kv/store.go
xichen2020: nit: make this a `errNilValidator = errors.New("no validator set on StoreOptions so validation is not applicable")`

service/r2/service.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/instrument"
 
+	"github.com/pborman/uuid"
```
xichen2020: Reorder imports.

service/r2/service.go
```diff
@@ -66,6 +69,15 @@ var (
 	errNilRequest = errors.New("Nil request")
 )
 
+type endpoint struct {
```
xichen2020: nit: perhaps rename this to `route`?

service/r2/service.go
```diff
@@ -66,6 +69,15 @@ var (
...
+	path   string
+	method string
+}
+
+var authorizationRegistry = map[endpoint]auth.AuthorizationType{
+	{path: validateRuleSetPath, method: http.MethodPost}: auth.AuthorizationTypeReadOnly,
```
xichen2020: nit: add a comment here saying validation only requires read access?

service/r2/service.go
```diff
@@ -155,6 +167,32 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
 	}
 }
 
+func getDefaultAuthorizationTypeForHTTPMethod(method string) (auth.AuthorizationType, error) {
```
xichen2020: nit: remove the `get` prefix to be more idiomatic in Go.

service/r2/service.go
```diff
@@ -155,6 +167,32 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
...
+	switch method {
+	case http.MethodGet:
+		return auth.AuthorizationTypeReadOnly, nil
+	case http.MethodPost, http.MethodPut, http.MethodDelete:
```
xichen2020: The old logic has `Patch` in here too, do we no longer need it?

service/r2/service.go
```diff
@@ -155,6 +167,32 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
...
+		return auth.AuthorizationTypeReadWrite, nil
+	default:
+		return auth.AuthorizationTypeNone, fmt.Errorf("unsupported http method %s for getting authorization type", method)
```
xichen2020: nit: `unsupported HTTP method...`

auth/auth.go
```diff
@@ -27,20 +27,34 @@ import (
...
+
 type errorResponseHandler func(w http.ResponseWriter, code int, msg string) error
 
 const (
 	// UserIDField is a key
 	UserIDField keyType = iota
 )
 
+const (
+	// AuthorizationTypeNone is the no authorizationType case.
```
xichen2020: Perhaps we should also add a `Unknown` here too?

service/r2/service.go
```diff
@@ -155,6 +167,32 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
...
+
+func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2HandlerFunc) {
+	authType, exists := authorizationRegistry[endpoint{path: path, method: method}]
+	if !exists {
+		var err error
+		authType, err = getDefaultAuthorizationTypeForHTTPMethod(method)
+		if err != nil {
+			// Panic if cannot find an authorization type for the route and/or method. This indicates an
+			// unrecognized route and service should panic on startup when registering routes.
+			panic(err)
```
xichen2020: Hm we should generally return error and not panic.

service/r2/service.go
```diff
@@ -155,6 +167,32 @@ func (h r2Handler) handleError(w http.ResponseWriter, opError error) {
...
+	default:
+		return auth.AuthorizationTypeNone, fmt.Errorf("unsupported http method %s for getting authorization type", method)
+	}
+}
+
+func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2HandlerFunc) {
+	authType, exists := authorizationRegistry[endpoint{path: path, method: method}]
+	if !exists {
+		var err error
+		authType, err = getDefaultAuthorizationTypeForHTTPMethod(method)
```
xichen2020: nit: `if authType, err = ...; err != nil { return err }`

service/r2/service.go
```diff
@@ -187,35 +225,35 @@ func (s *service) URLPrefix() string { return s.rootPrefix }
...
 
 	// Mapping Rule actions
-	router.Handle(mappingRuleRoot, h.wrap(s.createMappingRule)).Methods(http.MethodPost)
+	registerRoute(router, mappingRuleRoot, http.MethodPost, h, s.createMappingRule)
 
-	router.Handle(mappingRuleWithIDPath, h.wrap(s.fetchMappingRule)).Methods(http.MethodGet)
-	router.Handle(mappingRuleWithIDPath, h.wrap(s.updateMappingRule)).Methods(http.MethodPut, http.MethodPatch)
-	router.Handle(mappingRuleWithIDPath, h.wrap(s.deleteMappingRule)).Methods(http.MethodDelete)
+	registerRoute(router, mappingRuleWithIDPath, http.MethodGet, h, s.fetchMappingRule)
+	registerRoute(router, mappingRuleWithIDPath, http.MethodPut, h, s.updateMappingRule)
```
xichen2020: No patch now? If so, do we need to update swagger doc too?

service/r2/service.go
```diff
@@ -187,35 +225,35 @@ func (s *service) URLPrefix() string { return s.rootPrefix }
...
 
 	// Rollup Rule actions
-	router.Handle(rollupRuleRoot, h.wrap(s.createRollupRule)).Methods(http.MethodPost)
+	registerRoute(router, rollupRuleRoot, http.MethodPost, h, s.createRollupRule)
 
-	router.Handle(rollupRuleWithIDPath, h.wrap(s.fetchRollupRule)).Methods(http.MethodGet)
-	router.Handle(rollupRuleWithIDPath, h.wrap(s.updateRollupRule)).Methods(http.MethodPut, http.MethodPatch)
-	router.Handle(rollupRuleWithIDPath, h.wrap(s.deleteRollupRule)).Methods(http.MethodDelete)
+	registerRoute(router, rollupRuleWithIDPath, http.MethodGet, h, s.fetchRollupRule)
+	registerRoute(router, rollupRuleWithIDPath, http.MethodPut, h, s.updateRollupRule)
```
xichen2020: Same question as above.

service/r2/service.go
```diff
@@ -553,3 +600,49 @@ type ruleSetJSON struct {
 	MappingRules  []mappingRuleJSON `json:"mappingRules"`
 	RollupRules   []rollupRuleJSON  `json:"rollupRules"`
 }
+
+type genID int
```
xichen2020: nit: `idGenType` might be better.

service/r2/service.go
```diff
@@ -553,3 +600,49 @@ type ruleSetJSON struct {
...
+
+const (
+	genIDFalse genID = iota
```
xichen2020: `generateID`

service/r2/service.go
```diff
@@ -553,3 +600,49 @@ type ruleSetJSON struct {
...
+	genIDTrue
```
xichen2020: `dontGenerateID`.

service/r2/service.go
```diff
@@ -553,3 +600,49 @@ type ruleSetJSON struct {
...
+)
+
+// ruleSetSnapshot create a RuleSetSnapshot from a rulesetJSON. If the ruleSetJSON has no IDs
+// for any of its mapping rules or rollup rules, it generates missing IDs and sets as a string UUID
+// string so they can be stored in a mapping (id -> rule).
+func (r ruleSetJSON) ruleSetSnapshot(genID genID) (*rules.RuleSetSnapshot, error) {
+
```
xichen2020: Remove blank line.

auth/auth.go
m-sandusky: Agreed that they're verbose but I'd argue for them based on clarity for a couple reasons:
1) Changing them to None/ReadOnly/etc reserves those names for the entire auth package
2) Due to the way Go does (or rather doesn't do) enums it's not clear that something like None or ReadOnly would be associated with an AuthorizationType

service/r2/service.go
m-sandusky: Sorry, need to fix my local config to not order these in weird ways.

service/r2/service.go
m-sandusky: I'll add it back in here for future use cases but I propose we don't use it currently.

service/r2/service.go
m-sandusky: I removed patch because looking at the Swagger docs we don't reference PATCH for any endpoint. I'd argue that in that case we should restrict the API to what's listed in the docs.

service/r2/convert.go
```diff
@@ -195,3 +198,49 @@ func newRuleSetJSON(latest *rules.RuleSetSnapshot) ruleSetJSON {
 		RollupRules:   rrJSON,
 	}
 }
+
+// TODO(@SANDUSKY) MOVE THIS
+type idGenType int
+
+const (
+	generateID idGenType = iota
```
jskelcy: NIT: Given that this is a flag, and probably always will be, this name is a little misleading. Making it sound like there could be more than 2 values.

service/health/service.go
```diff
@@ -61,10 +61,11 @@ func (s *service) URLPrefix() string {
 	return healthURL
 }
 
-func (s *service) RegisterHandlers(router *mux.Router) {
+func (s *service) RegisterHandlers(router *mux.Router) error {
```
jskelcy, line 64: What does this function need to return something now?

service/r2/service.go
```diff
@@ -96,6 +99,40 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
...
+		return auth.AuthorizationTypeUnknown
+	}
+}
+
+func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2HandlerFunc) error {
+	authType, exists := authorizationRegistry[route{path: path, method: method}]
+	if !exists {
+		var err error
+		if authType = defaultAuthorizationTypeForHTTPMethod(method); authType == auth.AuthorizationTypeUnknown {
+			return err
```
jskelcy: are we returning an empty error here?

jskelcy (approved):

auth/auth.go
xichen2020: That's fair, in that case perhaps call them `UnknownAuthorization`, `NoAuthorization`, `ReadOnlyAuthorization`, etc.?

service/r2/service.go
xichen2020: If we don't use it, let's not add it to the list.

service/r2/service.go
xichen2020: Totally agree, just wasn't sure if we were using PATCH in the endpoints.

auth/simple.go
```diff
@@ -92,23 +92,43 @@ type simpleAuthorization struct {
...
+			return err
+		}
+		return a.authorizeUserForWrite(userID)
 	default:
-		return fmt.Errorf("unsupported request method: %s", httpMethod)
+		return fmt.Errorf("unsupported authorization type %v passed to handler", authType)
+	}
+}
+
+func (a simpleAuthorization) authorizeUserForRead(userID string) error {
```
xichen2020: Can we pull the common code from `authorizeUserForRead` and `authorizeUserForWrite` into a common `authorizeUserForAccess` function and reuse it? The `authorizeUserForAccess` implementation can just be the same as the old `authorizeUser` function.

service/r2/convert.go
```diff
@@ -21,8 +21,11 @@
 package r2
 
 import (
+	"fmt"
+
 	"github.com/m3db/m3metrics/policy"
 	"github.com/m3db/m3metrics/rules"
+	"github.com/pborman/uuid"
```
xichen2020: Blank line between m3db imports and 3rd party imports.

service/r2/convert.go
```diff
@@ -195,3 +198,49 @@ func newRuleSetJSON(latest *rules.RuleSetSnapshot) ruleSetJSON {
 		RollupRules:   rrJSON,
 	}
 }
+
+// TODO(@SANDUSKY) MOVE THIS
```
xichen2020: Is this TODO still necessary?

service/r2/convert_test.go
```diff
@@ -0,0 +1,478 @@
+package r2
+
+import (
+	"testing"
+
+	"github.com/m3db/m3metrics/policy"
+	"github.com/m3db/m3metrics/rules"
```
xichen2020, line 7: Blank line between m3db imports and 3rd party imports.

service/r2/kv/store.go
```diff
@@ -37,6 +38,8 @@ type store struct {
 	updateHelper rules.RuleSetUpdateHelper
 }
 
+var errNilValidator = errors.New("no validator set on StoreOptions so validation is not applicable")
```
xichen2020, line 41: super nit: might read a little better to say "no validator set so validation is not applicable".

service/r2/kv/store.go
```diff
@@ -73,6 +76,15 @@ func (s *store) FetchNamespaces() (*rules.NamespacesView, error) {
 	}, nil
 }
 
+func (s *store) ValidateRuleSet(rs *rules.RuleSetSnapshot) error {
+	validator := s.opts.Validator()
+	// If no validator is set, then the validation functionality is not applicable
```
xichen2020: Comments should end w/ a period.

service/r2/service.go
```diff
@@ -96,6 +99,39 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
 	}
 }
 
+type route struct {
```
xichen2020: super nit: might wanna consider moving `route` to the bottom so the service related pieces are in one piece.

service/r2/service.go
```diff
@@ -96,6 +99,39 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
...
+	path   string
+	method string
+}
+
+var authorizationRegistry = map[route]auth.AuthorizationType{
+	// This validation route should only require read access.
+	{path: validateRuleSetPath, method: http.MethodPost}: auth.AuthorizationTypeReadOnly,
+}
+
+func defaultAuthorizationTypeForHTTPMethod(method string) auth.AuthorizationType {
```
xichen2020: I'd still return an error here just to make it very explicit that if we don't see one of the expected methods, it should be an error, and check error in `registerRoute`.

services/r2ctl/main/main.go
```diff
@@ -130,7 +130,10 @@ func main() {
 	})
 	httpServerInstrumentOpts := instrumentOpts.SetMetricsScope(httpServerScope)
 	httpServerOpts := cfg.HTTP.NewServerOptions(httpServerInstrumentOpts)
-	server := http.NewServer(listenAddr, httpServerOpts, r2Service, healthService)
+	server, err := http.NewServer(listenAddr, httpServerOpts, r2Service, healthService)
+	if err != nil {
+		logger.Fatalf("could not initialize new server: %v", err)
```
xichen2020: super nit: `initialize` -> `create`.

service/r2/service.go
```diff
@@ -126,38 +162,70 @@ func NewService(
...
 func (s *service) URLPrefix() string { return s.rootPrefix }
 
-func (s *service) RegisterHandlers(router *mux.Router) {
+func (s *service) RegisterHandlers(router *mux.Router) error {
 	// Namespaces action
 	h := r2Handler{s.logger, s.authService}
 
-	router.Handle(namespacePath, h.wrap(s.fetchNamespaces)).Methods(http.MethodGet)
-	router.Handle(namespacePath, h.wrap(s.createNamespace)).Methods(http.MethodPost)
+	// Namespaces actions
```
xichen2020: Might be more cleaner to do:
```
// Register routes.
routeWithHandlers := []struct{
   route route
   handler r2HandlerFnc
} {
  {route: {path: namespacePath: method: http.MethodGet}, handler: s.fetchNamespaces},
  {route: {path: namespacePath: method: http.MethodPost}, handler: s.createNamespace},
  ...
}
for _, rh := range routeWithHandlers {
  if err := registerRoute(router, rh.route.path, rh.route.method, h, rh.handler); err != nil {
    return err
  }
}
return nil
```

auth/simple.go
```diff
@@ -110,30 +110,27 @@ func (a simpleAuthorization) authorize(authType AuthorizationType, userID string
 	}
 }
 
-func (a simpleAuthorization) authorizeUserForRead(userID string) error {
-	if !a.readWhitelistEnabled {
-		return nil
-	}
-
-	for _, u := range a.readWhitelistedUserIDs {
+func authorizeUserForAccess(userID string, whitelistedUserIDs []string) error {
```
xichen2020: nit: could take a boolean parameter `enabled` and further extract the enabled check in the `read` and `write` methods below.

service/r2/service.go
```diff
@@ -99,32 +99,28 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
...
 }
 
 func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2HandlerFunc) error {
 	authType, exists := authorizationRegistry[route{path: path, method: method}]
 	if !exists {
-		if authType = defaultAuthorizationTypeForHTTPMethod(method); authType == auth.AuthorizationTypeUnknown {
-			return fmt.Errorf("unknown authorization type for method %s at path %s", method, path)
+		var err error
+		if authType, err = defaultAuthorizationTypeForHTTPMethod(method); err != nil {
+			return fmt.Errorf("could not register route for path %s and method %s, error: %v", method, path, err)
```
xichen2020: swap `method` and `path`?

m-sandusky: Thanks for the catch.

xichen2020:
LGTM, good stuff!

Might wanna wait for travis though.

xichen2020 (approved):
LGTM

PR #52, opened 2018-01-11
Log per request

cw9:
Log every http request

@xichen2020 @m-sandusky


service/r2/service.go
```diff
@@ -230,13 +235,27 @@ func (s *service) handleRoute(rf routeFunc, r *http.Request, namespace string) (
 	}
 	start := s.nowFn()
 	data, err := rf(s, r)
-	s.metrics.recordMetric(r.RequestURI, r.Method, namespace, time.Since(start), err)
+	t := time.Since(start)
```
xichen2020: Probably want to do s.nowFn().Sub(start) here in case `nowFn` is different from `time.Now`.

cw9: sounds good

xichen2020 (approved):
LGTM

PR #58, opened 2018-01-31
Properly propogate validation error

m-sandusky:
Fixes issue where validation endpoint would return 500 for invalid rulesets instead of 400.


xichen2020 (approved):
LGTM

PR #59, opened 2018-03-12
update and delete functions now take an optional rule-set-version

jskelcy:
m3ctrl right now relies on tradition CRUD semantics for interacting with rules. With the current implementation, we are not able to do a _check and set_ operation which are very powerful in a distributed systems setting. Right now the underlying storage layer of m3ctl is KV, which provides a check and set method for data with a given _version_. This diff updates the HTTP/JSON interface to accept a _rule-set-version_ on rule updates and deletes so that we take advantage of KV check and sets.

NOTE:

- Currently, if you query a single rule it does not return a _rule-set-version_. This means in order to update a rule and use check and set you would need to query the entire rule set, which does return a rule-set-version, then make your change with the version. I think it is reasonable to have those endpoints return a version number, however, it may be better to add those in a separate PR.

- In all cases, I have not made the _rule-set-version_ fields required. This will not break API backward compatibility. I am not sure the best way to go forward here. On one hand, we could say that if these fields are omitted then we skip the check and set, but that is potentially dangerous. The other options are to make it required and break back compat, and fix the UI, or create v2 endpoints. I am personally in the _create a new version_ camp, as it will be much less complicated than making it optional while making the API cleaner. The final solution is to break back compat but that should be avoided whenever possible.


public/r2/v1/swagger/swagger.json
```diff
@@ -387,7 +394,15 @@
                     "200": {
                         "description": "Rule updated",
                         "schema": {
-                            "$ref": "#/definitions/MappingRule"
+                            "type": "object",
+                            "properties": {
+                                "rollup-rule": {
```
m-sandusky: Should this be called mapping-rule?

public/r2/v1/swagger/swagger.json
```diff
@@ -442,7 +464,7 @@
                     "200": {
                         "description": "Ruleset updated",
                         "schema": {
-                            "$ref": "#/definitions/ApiResponse"
+                            "$ref": "#/definitions/DeleteResponse"
```
m-sandusky: A little confused about the DeleteResponse here - why do we need to return the version for this?

public/r2/v1/swagger/swagger.json
jskelcy: oops.

public/r2/v1/swagger/swagger.json
jskelcy: well, the version is for a rule set and this operation updates the current ruleset snapshot, there will be one less rule. If you then wanted to do further modifications after the delete you would be operating on a stale version of the rule set so it needs to be returned.

jskelcy:
Updated this PR with a different approach. I have added a bulk update endpoint where a set of changes to a ruleset can be submitted. There is a flag which will apply all the updates either atomically or not based on the atomic flag. Initially, only atomic operations will be supported. This endpoint accepts a rule-set-verison which will be used as a `check and set` version number for the ruleset. If the version is stale an error will be returned from the endpoint.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,79 @@
...
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                        "in": "path",
+                        "name": "namespaceID",
```
saminzadeh, line 203: This is probably unnecessary to be required. You are updating the resource which is in the URL, so this can be inferred.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,79 @@
...
+                        "description": "The updates to the ruleset",
+                        "required": true,
+                        "schema": {
+                        "$ref": "#/definitions/RuleSetChange"
+                        }
+                    },
+                    {
+                        "in": "body",
+                        "name": "rule-set-version",
+                        "description": "The ruleset version to update, for check and set operations",
```
saminzadeh: If this isn't passed, will it skip this check-set validation? If so, let's make a comment in the description here.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,79 @@
...
+                            "$ref": "#/definitions/ApiResponse"
+                        }
+                    },
+                    "404": {
+                        "description": "No such namespace",
+                        "schema": {
+                            "$ref": "#/definitions/ApiResponse"
+                        }
+                    },
+                    "409": {
```
saminzadeh, line 238: When would this occur? This is a PUT. This implies that we are creating something via a POST

public/r2/v1/swagger/swagger.json
jskelcy: Interesting. I took a look at the other endpoint URIs and it looked like they required parameters that were part of the URL. Perhaps we should keep if for consistency.

public/r2/v1/swagger/swagger.json
jskelcy: This should actually be required. I was thinking that if you did things non-atomically you would not need to pass this, but I was wrong about that, but it should always be required.

public/r2/v1/swagger/swagger.json
jskelcy: A 409 is the case when you provide a rule-set-version that is stale.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,79 @@
...
+                    {
+                        "in": "body",
+                        "name": "rule-set-version",
+                        "description": "The ruleset version to update, for check and set operations",
+                        "required": false,
+                        "type": "integer"
+                    },
+                    {
+                        "in": "body",
+                        "name": "atomic",
```
xichen2020: Actually on a second thought, we probably don't need this to begin with and always require a ruleset version to start with. We can always relax that constraint later if needed.

public/r2/v1/swagger/swagger.json
```diff
@@ -863,20 +936,80 @@
                     "type": "integer",
                     "format": "unixMillis"
                 },
-                "mappingRules": {
+                "mappingRulesChanges": {
```
xichen2020: I think these should be mapping rules instead of mapping rule changes?

public/r2/v1/swagger/swagger.json
```diff
@@ -863,20 +936,80 @@
...
                     "type": "array",
                     "items": {
                         "$ref": "#/definitions/MappingRule"
                     }
                 },
-                "rollupRules": {
+                "rollupRulesChanges": {
```
xichen2020: I think these should be rollup rules instead of rollup rule changes?

public/r2/v1/swagger/swagger.json
saminzadeh: If thats the case, then lets keep it consistent

public/r2/v1/swagger/swagger.json
saminzadeh: Should the description clarify this?

jskelcy: I see. The description here is a little confusing. But I think a 409 is still relevant.

public/r2/v1/swagger/swagger.json
jskelcy: I should have caught this. I think this was bad find and replace.

saminzadeh (approved):

public/r2/v1/swagger/swagger.json
```diff
@@ -84,7 +84,7 @@
                         }
                     },
                     "409": {
-                        "description": "Namespace already exists",
+                        "description": "The ruleset got updated while you were looking at it.",
```
xichen2020: Hm I thought we'd return 409 when the namespace already existed? If someone calls this API programmatically without checking whether the namespace exists beforehand, they can get 409 if they try to create a namespace that exists. If so, probably wanna keep it as "Namespace already exists"?

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
                 }
             }
         },
+        "/namespaces/{namespaceID}/ruleset/bulk": {
+            "put": {
```
xichen2020: Should this be POST rather than PUT since calling this endpoint twice with the same set of changes are not idempotent?

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
...
+                "tags":[
+                    "namespaces"
+                ],
+                "summary": "Bulk update endpoint for namespace changes",
```
xichen2020: nit: `for ruleset changes`.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
...
+                        "name": "ruleset-change",
+                        "description": "The updates to the ruleset",
+                        "required": true,
+                        "schema": {
+                        "$ref": "#/definitions/RuleSetChange"
+                        }
+                    },
+                    {
+                        "in": "body",
+                        "name": "rule-set-version",
```
xichen2020: Should this be `ruleset-version` instead of `rule-set-version`?

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
...
+                        "description": "The ruleset version to update, for check and set operations",
```
xichen2020: nit: might be more clear to say "The ruleset version to atomically apply bulk updates against"

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
...
+                    {
+                        "in": "path",
+                        "name": "namespaceID",
+                        "description": "The name of the namespace",
+                        "type": "string",
+                        "required": true
+                    },
+                    {
+                        "in": "body",
+                        "name": "ruleset-change",
```
xichen2020: This captures the set of rule changes yeah? If so might be better to call this `ruleset-changes`?

public/r2/v1/swagger/swagger.json
```diff
@@ -184,6 +184,72 @@
...
+                        "description": "The updates to the ruleset",
+                        "required": true,
+                        "schema": {
+                        "$ref": "#/definitions/RuleSetChange"
```
xichen2020: Should this be `RuleSetChanges`?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
                 }
             }
         },
+        "RuleSetChange": {
```
xichen2020: `RuleSetChanges`?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+                    "type": "string",
+                    "enum": ["add", "remove", "change"],
+                    "description": "type of operation on rule set"
+                },
+                "MappingRuleChanges": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/MappingRuleChange"
+                    },
+                    "description": "list of mapping rule diffs"
```
xichen2020: "list of mapping rule changes"

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+                },
+                "RollupRuleChanges": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/RollupRuleChange"
+                    },
+                    "description": "list of rollup rule diffs"
```
xichen2020: "list of rollup rule changes"

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
                 }
             }
         },
+        "RuleSetChange": {
+            "type": "object",
+            "properties": {
+                "operation": {
+                    "type": "string",
+                    "enum": ["add", "remove", "change"],
```
xichen2020: Does this mean we might add / delete namespace as part of the bulk operation? Might wanna think about if we want to allow that given that we already have APIs for adding new namespaces and deleting existing namespaces?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+                    "description": "type of operation on rule"
+                },
+                "from": {
+                    "$ref": "#/definitions/MappingRuleData",
+                    "description": "old version of rule"
```
xichen2020: Perhaps better to say "old rule data"? 

Also this can be null for "add" and "change" operations I assume?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+                },
+                "to": {
+                    "$ref": "#/definitions/MappingRuleData",
+                    "description": "new version of rule"
```
xichen2020: Perhaps better to say "new rule data"?

Also this can be null for "remove" operations I assume?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+            "type": "object",
+            "properties": {
+                "operation": {
+                    "type": "string",
+                    "enum": ["add", "remove", "change"],
+                    "description": "type of operation on rule"
+                },
+                "from": {
+                    "$ref": "#/definitions/RollupRuleData",
+                    "description": "old version of rule"
```
xichen2020: Perhaps better to say "old rule data"?

Also this can be null for "add" and "change" operations I assume?

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,66 @@
...
+                },
+                "to": {
+                    "$ref": "#/definitions/RollupRuleData",
+                    "description": "new version of rule"
```
xichen2020: Perhaps better to say "new rule data"?

Also this can be null for "remove" operations I assume?

public/r2/v1/swagger/swagger.json
jskelcy: This was something I was thinking about. It does seem like this endpoint would be more simple if we did not allow for these types of operations, but if users are going to be hitting this endpoint programmatically it might be easier to have a one-stop endpoint for manipulating rules. I can remove it for now and if we need to add it back later we can.

xichen2020 (approved):
LGTM.

public/r2/v1/swagger/swagger.json
```diff
@@ -877,6 +943,61 @@
...
+                    "description": "list of rollup rule changes"
+                }
+            }
+        },
+        "MappingRuleChange": {
+            "type": "object",
+            "properties": {
+                "operation": {
+                    "type": "string",
+                    "enum": ["add", "remove", "change"],
```
xichen2020: Can we please change `remove` to `delete` for consistency? That's what we use in m3metrics as well.

PR #61, opened 2018-03-27
update to the latest m3metrics

jskelcy:
This diff does an update of a number of the deps of m3ctl.

IN THIS PR:
- All the `View` structs changed import paths to `models.XXViews`

- M3Cluster, M3Metrics, Protobuf and Net packages were upgraded to a new hash. Only M3Metrics was intended however the others were needed for compatibility.

- Mock, etcd and grpc were updated to use semver versions. Mock uses a fuzzy version consistent with other M3 repos.

- ` clusterkv.Configuration` is not longer an exported struct so we have migrated to ` clusterkv.OverrideConfiguration`. These appear to be functionally equivalent, just a name change.

NOT IN THIS PR:

- M3Metrics now has an exported JSON serializable rule type. This is equivalent to the type in the M3Ctl repo, in subsequent PRs the migration will happen.


glide.yaml
```diff
@@ -26,9 +26,9 @@ import:
 - package: github.com/go-ole/go-ole
   version: de8695c8edbf8236f30d6e1376e20b198a028d42
 - package: github.com/golang/mock
-  version: bd3c8e81be01eef76d4b503f5e687d2d1354d2d9
+  version: ^1
```
m-sandusky, line 29: Nit: You already mentioned fuzzy versioning consistent with other M3 repos - was just wondering the rationale behind it. Is it so to not have to download multiple versions of what should be compatible versions of the library?

m-sandusky (approved):
Had a comment. LGTM in general though.

service/r2/convert.go
```diff
@@ -34,7 +34,7 @@ type namespaceJSON struct {
 	ForRuleSetVersion int    `json:"forRuleSetVersion"`
 }
 
-func newNamespaceJSON(nv *rules.NamespaceView) namespaceJSON {
+func newNamespaceJSON(nv *models.NamespaceView) namespaceJSON {
```
xichen2020, line 37: Do we need these ctors etc? Should these be replaced by what we defined in the `models` package in `m3metrics`?

service/r2/routes_test.go
```diff
@@ -204,64 +204,64 @@ func newMockStore() Store {
 	return mockStore{}
 }
 
-func (s mockStore) FetchNamespaces() (*rules.NamespacesView, error) {
-	return &rules.NamespacesView{}, nil
+func (s mockStore) FetchNamespaces() (*models.NamespacesView, error) {
```
xichen2020, line 207: Can you please create an issue so we can tackle using mockgen to generate mocks for m3ctl to replace the hand-written mocks?

service/r2/convert.go
jskelcy: Yeah that needs to be done. I mentioned in the PR description that I am going to do that as a separate PR as there are a few things that are blocked by this being landed and if we land this first than the things that are blocked and using the new structs can be worked on in parallel.

glide.yaml
jskelcy: I am not sure what you are asking here. As far as I know, only one version of a package is installed with glide. The rationale for this version is just consistency.

service/r2/routes_test.go
jskelcy: https://github.com/m3db/m3ctl/issues/62

👍

service/r2/convert.go
xichen2020: Cool sounds good.

service/r2/routes_test.go
xichen2020: Sweet thanks.

xichen2020 (approved):
LGTM

PR #63, opened 2018-03-28
use models from m3metrics rather than local json structs

jskelcy:
Much of the convert.go file that is in m3ctl has been moved to m3metrics. After updating the version of m3metrics we are now removing duplicate code.

This PR does not functionally change any code. Just a reorganization to remove duplicate code where possible.


xichen2020:
What happened to the 7+% coverage drop?

service/r2/models.go
```diff
@@ -0,0 +1,49 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package r2
+
+import (
+	"github.com/m3db/m3metrics/rules/models"
+)
+
+type mappingRuleHistoryJSON struct {
```
xichen2020: Can we move this to `m3metrics`? Feels strange to have these two data models in `m3ctl` and the rest of the data models in `m3metrics`.

service/r2/routes.go
```diff
@@ -212,16 +214,16 @@ func createRollupRule(s *service, r *http.Request) (data interface{}, err error)
 		return nil, err
 	}
 
-	rr, err := s.store.CreateRollupRule(namespaceID, rrj.rollupRuleView(), uOpts)
+	_, err = s.store.CreateRollupRule(namespaceID, rrj.ToRollupRuleView(), uOpts)
 	if err != nil {
 		return nil, err
 	}
-	return newRollupRuleJSON(rr), nil
+	return rrj, nil
```
xichen2020: Hm why the change here?

service/r2/models.go
jskelcy: Yeah I can move these over. I am not sure why we didnt do it initially. Either way I dont think it needs to block this diff.

service/r2/routes.go
jskelcy: I was thinking that if we already allocated the exact same object when deserializing the body of the request we dont need to do it again with the view that comes back from the store package. It might be a premature optimization, I can get rid of it.

jskelcy:
Looking into the code coverage. It appears that the file that was deleted was well covered so by removing the total coverage for the project dropped.

service/r2/models.go
jskelcy: Here is the diff to move them into m3metrics

https://github.com/m3db/m3metrics/pull/142

xichen2020 (approved):
LGTM

PR #64, opened 2018-04-03
Add endpoint for accepting ruleset changes

jskelcy:
Added bulk endpoint to change an entire ruleset at once. Also did refactoring where needed.


glide.yaml
```diff
@@ -56,7 +56,7 @@ import:
 - package: golang.org/x/net
   version: ab5485076ff3407ad2d02db054635913f017b0ed
 - package: github.com/gorilla/mux
-  version: 18fca31550181693b3a834a15b74b564b3605876
+  version: 1.6.1
```
jskelcy, line 59: We were on an old version which was missing a number of testing features so I brought up us to the latest tag.

service/r2/store/stub/store.go
```diff
@@ -21,11 +21,14 @@
 package stub
```
jskelcy, line 21: This implementation appears to be largely used for testing. New functions will not be implemented in this package but instead should use the mock storage implementation.

public/r2/v1/swagger/swagger.json
```diff
@@ -198,28 +198,29 @@
                     "application/json"
```
jskelcy, line 198: There were all kinds of swagger compilation problems with this file so I fixed them.

service/r2/requests.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
jskelcy: We did not have a great place to put request objects. Most of what we are using comes from m3metrics, but I think in this case this struct definition does not really belong there. I created this file for now but I am open to putting this somewhere else.

public/r2/v1/swagger/swagger.json
```diff
@@ -198,28 +198,29 @@
...
+                            }
                         }
                     },
                     {
-                        "in": "body",
-                        "name": "ruleset-version",
-                        "description": "The ruleset version to atomically apply bulk updates against",
-                        "required": true,
-                        "type": "integer"
+                        "in": "path",
```
m-sandusky, line 218: Is this a duplicate of the first element in this list?

service/r2/requests.go
```diff
@@ -0,0 +1,28 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package r2
+
+import "github.com/m3db/m3metrics/rules/models/changes"
+
+type bulkRequest struct {
+	RuleSetChanges changes.RuleSetChanges `json:"ruleset-changes"`
```
m-sandusky: Nit: JSON fields should follow camelCase convention?

service/r2/routes.go
```diff
@@ -40,7 +44,7 @@ func fetchNamespaces(s *service, _ *http.Request) (data interface{}, err error)
 
 func fetchNamespace(s *service, r *http.Request) (data interface{}, err error) {
 	vars := mux.Vars(r)
-	rs, err := s.store.FetchRuleSet(vars[namespaceIDVar])
+	rs, err := s.store.FetchRuleSetSnapshot(vars[namespaceIDVar])
```
m-sandusky, line 43: Not your change - but do we validate vars[namespaceIDVar] anywhere? Is it possible that this could cause issues if not validated/checked?

service/r2/requests.go
```diff
@@ -0,0 +1,28 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package r2
+
+import "github.com/m3db/m3metrics/rules/models/changes"
+
+type bulkRequest struct {
```
m-sandusky: Understand that we're consolidating endpoints and whatnot but is bulkRequest too generic of a name? It may be good to specify as bulkRuleSetRequest or something similar.

glide.yaml
xichen2020: Do you want to use `^1.6.1` instead of pinning it to a single version?

service/r2/mocks/storage_mock.go
```diff
@@ -0,0 +1,242 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
xichen2020: Can you set up the build script so the mocks can be auto-generated via mock-gen? This is pretty standard practice in other repos btw, so next time you change the `Store` interface, you just need to run `make mock-gen`.

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+	}
+	if req.RuleSetVersion != originalRS.Version() {
+		return nil, NewConflictError(
+			fmt.Sprintf("can not update namespace %s, version mismatch", vars[namespaceIDVar]),
+		)
+	}
+
+	newRS, err := applyChangesToRuleSet(
+		req.RuleSetChanges,
+		originalRS.ToMutableRuleSet().Clone(),
```
m-sandusky: Might be worth cloning the original mutable ruleset within this function. This way callers don't have to remember to clone the ruleset and accidentally mutating the original rule set. I'm also a fan of functions that don't mutate the inputs.

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+		uOpts,
+		s.updateHelper,
+	)
+	if err != nil {
+		return nil, NewBadInputError(err.Error())
```
m-sandusky: Nit/Just a passing comment: What's the original error type that is bubbled up? Is it worth discarding this information?

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+
+	rs, err := s.store.UpdateRuleSet(newRS)
+	if err != nil {
+		return nil, err
+	}
+	latest, err := rs.Latest()
+	if err != nil {
+		return nil, err
+	}
+	serializableRuleSet := models.NewRuleSet(latest)
```
m-sandusky: nit: Can just return this rather than init a new variable

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+
+	return serializableRuleSet, nil
+}
+
+func applyChangesToRuleSet(
```
m-sandusky: As mentioned previously, it might be worth taking a RuleSetSnapshot and then doing the .ToMutable().Clone() inside here.

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+	rsc changes.RuleSetChanges,
+	ruleSet rules.MutableRuleSet,
+	uOpts store.UpdateOptions,
+	helper rules.RuleSetUpdateHelper,
+) (rules.MutableRuleSet, error) {
+
+	if len(rsc.MappingRuleChanges) == 0 &&
```
m-sandusky: It might be worth adding a method to changes.RuleSetChanges that returns a bool that indicates whether there are any changes present. If the ruleset structure changes (and I think it may with the multi rollup rule stuff), it'd be good to have this logic contained in one place rather than updating different areas. It's also cleaner IMO.

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+		len(rsc.RollupRuleChanges) == 0 {
+		return nil, fmt.Errorf("invalid request: no ruleset changes detected")
+	}
+
+	meta := helper.NewUpdateMetadata(
+		time.Now().Unix(),
```
m-sandusky: Nit/just a thought: Interesting that the caller is responsible for setting the time. Not sure how this is used but would expect the receiver to set this.

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+		uOpts.Author(),
+	)
+
+	for _, mrChange := range rsc.MappingRuleChanges {
+		if mrChange.Op == "" {
+			return nil, fmt.Errorf("invalid request: changes must contain an op")
```
m-sandusky: nit: specify that this is MappingRule in the err string and/or create a new: 
`var errNoOpChage = errors.New(...)`

Also don't need fmt.Errorf

service/r2/routes.go
```diff
@@ -271,3 +275,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+			err := ruleSet.DeleteMappingRule(*mrChange.RuleID, meta)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	for _, rrChange := range rsc.RollupRuleChanges {
+		if rrChange.Op == "" {
+			return nil, fmt.Errorf("invalid request: changes must contain an op")
```
m-sandusky: Same comment as above

service/r2/service.go
```diff
@@ -45,8 +48,9 @@ const (
 )
 
 var (
-	namespacePrefix     = fmt.Sprintf("%s/{%s}", namespacePath, namespaceIDVar)
-	validateRuleSetPath = fmt.Sprintf("%s/{%s}/ruleset/validate", namespacePath, namespaceIDVar)
+	namespacePrefix       = fmt.Sprintf("%s/{%s}", namespacePath, namespaceIDVar)
+	validateRuleSetPath   = fmt.Sprintf("%s/{%s}/ruleset/validate", namespacePath, namespaceIDVar)
+	bulkUpdateRuleSetPath = fmt.Sprintf("%s/{%s}/ruleset/bulk", namespacePath, namespaceIDVar)
```
m-sandusky: Since we are only updating one rule set in this endpoint - would it be worth just dropping the bulk altogether and calling it updateRuleSetPath? Wondering if we need to distinguish bulk - my first thought is that bulk does multiple ops (which we technically do I guess) but from a consumer side it just looks like we're updating an entire ruleset.

public/r2/v1/swagger/swagger.json
jskelcy: Hmm I dont think so. It may be the case that github is just making this hard to look at. What is actually happening here is that there were two entries for where `in: body` was present, which is not valid swagger. Also the body params must come first, so I moved the body params to the top of the list and merged the two elements in the list togther. I would suggest taking a look [here](https://github.com/m3db/m3ctl/blob/5a1720261bfcba373fd35fc088728e4c699d6b5c/public/r2/v1/swagger/swagger.json), it might be easier to see what is going on.

service/r2/requests.go
jskelcy: yeah I think there are some inconsistencies in the existing `swagger.json` and I was trying to stick to the repo convention. But the more I look the more I see that even that file is all over the place so Ill make the camel case.

service/r2/routes.go
jskelcy: thats a great point. I think in this case this is a read-only endpoint so no one can mess with data in the database, and there is already some auth validation. But this is worth checking in on. I am not sure it is within the scope of this diff but Ill take a look and if it looks like something needs to be done here I can create a ticket.

service/r2/requests.go
jskelcy: 👍

service/r2/routes.go
jskelcy: 👍

service/r2/routes.go
jskelcy: Potentially. I will have to think about this one because even if a ruleset structure changes its not 100% that the change object will change very much.

service/r2/routes.go
jskelcy: Yeah that's a great idea. Initially, there was some formatting in this error messag but it was taken out because there was potential for a panic. Now that there is no formatting the pattern you are suggesting make more sense.

service/r2/routes.go
jskelcy: So in this scenario m3ctl is the receiver, or at least the last thing that touches the data before it goes in a DB. There are some DBs that offer timestamp fields which can be updated everytime the data is accessed but because KV is storing binary blobs that is not the case here.

service/r2/service.go
jskelcy: yeah I am with that. Ill changes the path too. I could see how it might be misleading.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,42 +184,42 @@
...
         },
-        "/namespaces/{namespaceID}/ruleset/bulk": {
+        "/namespaces/{namespaceID}/ruleset/update": {
             "post": {
                 "tags":[
                     "namespaces"
                 ],
-                "summary": "Bulk update endpoint for ruleset changes",
-                "operationId": "bulkUpdateRuleSet",
+                "summary": "Update endpoint for a ruleset, allowing for multiple multiple atomic changes to a ruleset at once.",
```
m-sandusky: Delete second 'multiple'
Also super nit: for a public API mentioning multiple atomic changes could be confusing. Perhaps allowing for atomic changes for an entire RuleSet

public/r2/v1/swagger/swagger.json
```diff
@@ -184,42 +184,42 @@
...
+                "operationId": "udateRuleSet",
```
m-sandusky: sp: updateRuleSet

public/r2/v1/swagger/swagger.json
m-sandusky: Ah my bad - mislead by the UI

service/r2/routes.go
m-sandusky: Yeah fair enough. Also somewhat related, there;s also the case that namespaceIDVar is empty and we should check if v, ok := vars[key], where ok is true? Not a big deal though since it'll just look up "" in the worst case and send back the appropriate error.

service/r2/routes.go
m-sandusky: ping (in case you didn't see)

service/r2/routes.go
```diff
@@ -271,3 +280,117 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+	}
+	if req.RuleSetVersion != originalRS.Version() {
+		return nil, NewConflictError(
+			fmt.Sprintf("can not update namespace %s, version mismatch", vars[namespaceIDVar]),
+		)
+	}
+
+	newRS, err := applyChangesToRuleSet(
+		req.RuleSetChanges,
+		originalRS.ToMutableRuleSet(),
```
m-sandusky: I'd go even further and say that the originalRuleSet should be passed in and then the proper conversions can happen within the function.

jskelcy: Interesting idea but then the function would take in a `rules.Ruleset` and return a `rules.MutableRuleSet`. I was worried that it may be a little unclear but ill try it out.

service/r2/routes.go
jskelcy: Yeah I think personally I would leave the empty check as is. There is already checks in place further down the call stack that will detect this as a problem and raise the correct errors which will result if a 404. Not sure we need to add another check here.

service/r2/service.go
```diff
@@ -128,29 +134,32 @@ func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2Ha
...
 
 // NewService creates a new r2 service using a given store.
 func NewService(
 	rootPrefix string,
 	authService auth.HTTPAuthService,
-	store Store,
+	store store.Store,
 	iOpts instrument.Options,
 	clockOpts clock.Options,
+	rulePropigationDelay time.Duration,
```
m-sandusky: Spelling: propagation

service/r2/service.go
```diff
@@ -128,29 +134,32 @@ func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2Ha
...
-	logger      log.Logger
-	nowFn       clock.NowFn
-	metrics     serviceMetrics
+	rootPrefix   string
+	store        store.Store
+	authService  auth.HTTPAuthService
+	logger       log.Logger
+	nowFn        clock.NowFn
+	metrics      serviceMetrics
+	updateHelper rules.RuleSetUpdateHelper
```
m-sandusky: Is this a helper or just update options?

service/r2/service.go
m-sandusky: propagation

service/r2/service.go
```diff
@@ -128,29 +134,32 @@ func registerRoute(router *mux.Router, path, method string, h r2Handler, hf r2Ha
...
-		logger:      iOpts.Logger(),
-		nowFn:       clockOpts.NowFn(),
-		metrics:     newServiceMetrics(iOpts.MetricsScope(), iOpts.MetricsSamplingRate()),
+		rootPrefix:   rootPrefix,
+		store:        store,
+		authService:  authService,
+		logger:       iOpts.Logger(),
+		nowFn:        clockOpts.NowFn(),
+		metrics:      newServiceMetrics(iOpts.MetricsScope(), iOpts.MetricsSamplingRate()),
+		updateHelper: rules.NewRuleSetUpdateHelper(rulePropigationDelay),
```
m-sandusky: propagation

service/r2/service.go
jskelcy: It is a helper. It appears to be this factory sturct that comes from m3metircs.

m-sandusky (approved):

Makefile
```diff
@@ -18,6 +18,12 @@ metalint_exclude    := .excludemetalint
 m3ctl_package       := github.com/m3db/m3ctl
 gopath_prefix       := $(GOPATH)/src
 vendor_prefix       := vendor
+mocks_output_dir 		:= generated/mocks/mocks
```
xichen2020: These should be aligned with the rest of the assignment.

Makefile
```diff
@@ -102,4 +108,25 @@ clean:
 all: lint metalint test-ci-unit services build-ui
 	@echo Made all successfully
 
+.PHONY: install-licence-bin
+install-license-bin: install-vendor
+	@echo Installing node modules
+	[ -d $(license_node_modules) ] || (cd $(license_dir) && npm install)
+
```
xichen2020: Remove empty line here.

Makefile
```diff
@@ -102,4 +108,25 @@ clean:
 all: lint metalint test-ci-unit services build-ui
 	@echo Made all successfully
 
+.PHONY: install-licence-bin
+install-license-bin: install-vendor
+	@echo Installing node modules
```
xichen2020, line 113: I think this should be
```
.PHONY: install-licence-bin
install-license-bin: install-vendor
	@echo Installing node modules
	git submodule update --init --recursive
	[ -d $(license_node_modules) ] || (cd $(license_dir) && npm install)
```
?

generated/mocks/generate.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// mockgen rules for generating mocks using file mode
```
xichen2020: Add a period at the end.

glide.yaml
xichen2020: Ping

Makefile
jskelcy: Hmm this is super strange. It does not look like this locally. Ill figure out why this looks that way.

service/r2/routes.go
```diff
@@ -21,12 +21,21 @@
 package r2
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/gorilla/mux"
```
xichen2020: Import ordering.

glide.yaml
jskelcy: We discussed this offline and I said I would rather keep it pinned to a version.

service/r2/routes.go
```diff
@@ -21,12 +21,21 @@
...
+
+	"github.com/m3db/m3ctl/service/r2/store"
+	"github.com/m3db/m3metrics/rules"
 	"github.com/m3db/m3metrics/rules/models"
+	"github.com/m3db/m3metrics/rules/models/changes"
+)
 
-	"github.com/gorilla/mux"
+var (
+	errNoOpChange = errors.New("invalid request: changes must contain an op")
```
xichen2020: Perhaps `errMissingOpInChanges` is a better name?

service/r2/routes_test.go
```diff
@@ -23,11 +23,19 @@ package r2
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
```
xichen2020: Import ordering.

service/r2/routes_test.go
```diff
@@ -198,9 +567,125 @@ func newTestInstrumentMethodMetrics() instrument.MethodMetrics {
...
+	)
+	schema, _ := mrs.Schema()
+	rs, _ := rules.NewRuleSetFromSchema(version, schema, rules.NewOptions())
+	return rs
+}
+
 type mockStore struct{}
 
-func newMockStore() Store {
+func newMockStore() store.Store {
```
xichen2020: Do we still need this given that we have an auto-generated mock store?

Makefile
jskelcy: There was a sneaky tab. Ill update that.

services/r2ctl/main/main.go
```diff
@@ -108,12 +109,20 @@ func main() {
 		"service-name": "r2",
 	})
 	r2ServiceInstrumentOpts := instrumentOpts.SetMetricsScope(r2ServiceScope)
+	var propagationDelay time.Duration
```
xichen2020: Why do we need to set this here now?

services/r2ctl/config/r2ctl.go
```diff
@@ -53,14 +53,14 @@ type Configuration struct {
...
+	Store R2StoreConfiguration `yaml:"store"`
 
 	// Simple Auth Config.
 	Auth *auth.SimpleAuthConfig `yaml:"auth"`
 }
 
-// r2StoreConfiguration has all the fields necessary for an R2 store.
-type r2StoreConfiguration struct {
+// R2StoreConfiguration has all the fields necessary for an R2 store.
+type R2StoreConfiguration struct {
```
xichen2020: Why do we need to export this now?

service/r2/store/stub/store.go
```diff
@@ -326,7 +329,12 @@ func (s *store) DeleteNamespace(namespaceID string, uOpts r2.UpdateOptions) erro
 	}
 }
 
-func (s *store) FetchRuleSet(namespaceID string) (*models.RuleSetSnapshotView, error) {
+// This function is not supported. Use mocks package.
+func (s *store) FetchRuleSet(namespaceID string) (rules.RuleSet, error) {
+	return nil, errors.New("Do not use this function")
```
xichen2020: Perhaps declare
```
var errNotImplemented = errors.New("not implemented")
```
at the top of the file and use it here and in `UpdateRuleSet`?

service/r2/store/store.go
```diff
@@ -73,6 +77,9 @@ type Store interface {
 	// and rule ID.
 	FetchRollupRuleHistory(namespaceID, rollupRuleID string) ([]*models.RollupRuleView, error)
 
+	// UpdateRuleSet updates a ruleset with a given namespace
```
xichen2020: At a period at the end.

Makefile
jskelcy: O yeah. I was looking at the collector here:
https://github.com/m3db/m3collector/blob/master/Makefile#L46

But it looks like we are not consistent across m3 repos for example here:
https://github.com/m3db/m3metrics/blob/master/Makefile

I can put up a PR in the collector to fix this.

service/r2/service.go
```diff
@@ -359,13 +369,21 @@ func (s *service) fetchRollupRuleHistory(w http.ResponseWriter, r *http.Request)
 	return s.sendResponse(w, http.StatusOK, data)
 }
 
+func (s *service) updateRuleSet(w http.ResponseWriter, r *http.Request) error {
+	data, err := s.handleRoute(updateRuleSet, r, s.metrics.bulkUpdateRuleSet)
```
xichen2020: Do you still want to refer to the handler as `bulk...`?

service/r2/routes.go
```diff
@@ -271,3 +280,116 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+	if err != nil {
+		return nil, err
+	}
+	if req.RuleSetVersion != originalRS.Version() {
+		return nil, NewConflictError(
+			fmt.Sprintf("can not update namespace %s, version mismatch", vars[namespaceIDVar]),
+		)
+	}
+
+	newRS, err := applyChangesToRuleSet(
```
xichen2020: The convention here for all the handlers is to implement the logic that mutates ruleset in`m3metrics/rules/ruleset.go`, and call the corresponding function here. Can you please do that? You can use `UpdateMappingRule` as an example to see the pattern here.

service/r2/store/store.go
```diff
@@ -38,8 +39,11 @@ type Store interface {
 	// DeleteNamespace deletes the namespace for the given namespace ID.
 	DeleteNamespace(namespaceID string, uOpts UpdateOptions) error
 
-	// FetchRuleSet fetches the ruleset for the given namespace ID.
-	FetchRuleSet(namespaceID string) (*models.RuleSetSnapshotView, error)
+	// FetchRuleSet fetch ruleset from KV.
+	FetchRuleSet(namespaceID string) (rules.RuleSet, error)
```
xichen2020: I think if you move the rule mutation logic into `m3metrics`, you don't need this API.

service/r2/store/store.go
xichen2020: Also more importantly, the store should not directly mutate ruleset in this repo.

xichen2020 (changes requested):

service/r2/routes_test.go
jskelcy: yes. All the other tests still rely on it and this diff is already long. I will put up a follow up that migrates the tests to use the mocked version.

services/r2ctl/main/main.go
jskelcy: Prior to this diff we were not using the MutableRuleSet object to update entire rule sets. But it sounds like this is going to be moved to the m3metrics repo so this will eventually go away.

service/r2/service.go
jskelcy: Mark and I had a discussion earlier up in the thread, and parts of it were in person. But the idea was that bulk kind of made it sound like you could update multiple rulesets at once. This name seemed more clear.

service/r2/store/store.go
jskelcy: This is a wrapper interface that does some added validation, so while this repo does not do any direct manipulation this method must be here to wrap the underlying function.

service/r2/store/store.go
jskelcy: I think what you mean here is that rather than fetching, calling the apply method and writing there could be one method?

That could work, but all of the methods in this class follow the pattern of fetch -> mutate -> write. Perhaps this is worth changing the paradigm here but I am not sure it needs to happen before this diff.

service/r2/store/storage_mock.go
```diff
@@ -0,0 +1,244 @@
...
+// Source: github.com/m3db/m3ctl/service/r2/store/go
+
+package store
+
+import (
+	"github.com/m3db/m3metrics/rules"
+	"github.com/m3db/m3metrics/rules/models"
+	"github.com/m3db/m3metrics/rules/models/changes"
+
+	"github.com/golang/mock/gomock"
```
jskelcy: Not sure what the policy here is, but this did not order the imports in the way that we traditionally do. I think leaving it is the best course of action that way if someone generates mocks it will be a noop, if this is not what best practices dictate let me know.

service/r2/store/stub/store.go
```diff
@@ -21,12 +21,16 @@
...
 	"time"
 
 	"github.com/m3db/m3ctl/service/r2"
+	r2store "github.com/m3db/m3ctl/service/r2/store"
 	"github.com/m3db/m3metrics/policy"
+	"github.com/m3db/m3metrics/rules"
 	"github.com/m3db/m3metrics/rules/models"
+	"github.com/m3db/m3metrics/rules/models/changes"
 	"github.com/m3db/m3x/instrument"
 	"github.com/pborman/uuid"
```
jskelcy: Will fix import ordering.

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
+func (s *store) UpdateRuleSet(
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSet, error) {
+	rs, err := s.FetchRuleSet(rsChanges.Namespace)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	if version != rs.Version() {
```
jskelcy: This needs to be added here, despite using set and check in KV. This is because the user will pull in the rule set, apply the changes, and then send up the changes. Before we apply the changes locally to send to KV we must check for a version mismatch in case there has been after the user has sent up the proposed changes.

Makefile
```diff
@@ -18,6 +18,12 @@ metalint_exclude    := .excludemetalint
 m3ctl_package       := github.com/m3db/m3ctl
 gopath_prefix       := $(GOPATH)/src
 vendor_prefix       := vendor
+mocks_output_dir 	  := generated/mocks/mocks
```
xichen2020: Alignment.

service/r2/requests.go
xichen2020: Perhaps this belongs to `io.go` for general input and output related type defs and conversion logic if any?

service/r2/routes.go
```diff
@@ -271,3 +271,27 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
 	}
 	return models.NewRollupRuleSnapshots(hist), nil
 }
+
+func updateRuleSet(s *service, r *http.Request) (data interface{}, err error) {
```
xichen2020: nit: Can we please move this to below `validateRuleSet`? I tend to group similar methods together so they are easier to find.

service/r2/routes.go
```diff
@@ -271,3 +271,27 @@ func fetchRollupRuleHistory(s *service, r *http.Request) (data interface{}, err
...
+	var req updateRuleSetRequest
+	if err := parseRequest(&req, r.Body); err != nil {
+		return nil, NewBadInputError(err.Error())
+	}
+	uOpts, err := s.newUpdateOptions(r)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(req.RuleSetChanges.MappingRuleChanges) == 0 &&
```
xichen2020: Check this immediately after `parseRequest`? Usually better to perform as much validation as possible upfront.

service/r2/routes_test.go
xichen2020: Ping

service/r2/routes_test.go
```diff
@@ -198,9 +333,70 @@ func newTestInstrumentMethodMetrics() instrument.MethodMetrics {
 	return instrument.NewMethodMetrics(tally.NoopScope, "testRoute", 1.0)
 }
 
+func newBulkReqBody() updateRuleSetRequest {
```
xichen2020: Please add `test` to the method names in tests to distinguish them from normal non-test code, e.g., `newTestBulkReqBody`.

service/r2/routes_test.go
```diff
@@ -198,9 +333,70 @@ func newTestInstrumentMethodMetrics() instrument.MethodMetrics {
...
+						Name: "mappingRule3",
+					},
+				},
+			},
+		},
+	}
+}
+
+// nolint: unparam
+func newRuleSet(version int) rules.RuleSet {
```
xichen2020: Same as above, please do a sweep across the PR. Thanks.

service/r2/routes_test.go
xichen2020: Yeah sounds good, can you either create an issue or add a TODO here?

service/r2/service.go
```diff
@@ -359,13 +364,21 @@ func (s *service) fetchRollupRuleHistory(w http.ResponseWriter, r *http.Request)
 	return s.sendResponse(w, http.StatusOK, data)
 }
 
+func (s *service) updateRuleSet(w http.ResponseWriter, r *http.Request) error {
```
xichen2020: Can you move this to below `validateNamespace` to group similar methods together? Actually `validateNamespace` should probably be called `validateRuleSet` instead, can you update that as well (I know it's not your change but it looks strange :\)?

service/r2/service.go
```diff
@@ -75,6 +77,7 @@ type serviceMetrics struct {
 	updateRollupRule        instrument.MethodMetrics
 	deleteRollupRule        instrument.MethodMetrics
 	fetchRollupRuleHistory  instrument.MethodMetrics
+	bulkUpdateRuleSet       instrument.MethodMetrics
```
xichen2020: Perhaps we just call this `updateRuleSet` since you removed `bulk` from the swagger so we can be consistent to not include bulk in the method name?

service/r2/service.go
```diff
@@ -94,6 +97,7 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
 		updateRollupRule:        instrument.NewMethodMetrics(scope, "updateRollupRule", samplingRate),
 		deleteRollupRule:        instrument.NewMethodMetrics(scope, "deleteRollupRule", samplingRate),
 		fetchRollupRuleHistory:  instrument.NewMethodMetrics(scope, "fetchRollupRuleHistory", samplingRate),
+		bulkUpdateRuleSet:       instrument.NewMethodMetrics(scope, "bulkUpdateRuleSet", samplingRate),
```
xichen2020: Same here, perhaps change the method name to `updateRuleSet` as well.

service/r2/store/kv/store.go
```diff
@@ -24,7 +24,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/m3db/m3metrics/rules/models/changes"
```
xichen2020: Import ordering.

service/r2/store/kv/store_test.go
```diff
@@ -0,0 +1,313 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/m3db/m3metrics/rules"
+	"github.com/m3db/m3x/clock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/m3db/m3ctl/service/r2"
```
xichen2020: Import ordering.

service/r2/store/kv/store_test.go
```diff
@@ -0,0 +1,313 @@
...
+		}),
+	)
+	rulesStore := NewStore(mockedStore, storeOpts)
+	uOpts := r2store.NewUpdateOptions().SetAuthor("validUser")
+	_, err = rulesStore.UpdateRuleSet(rsChanges, 1, uOpts)
+	require.Error(t, err)
+	require.IsType(t, r2.NewConflictError(""), err)
+}
+
+func newRuleSetChanges(mrs models.MappingRuleViews, rrs models.RollupRuleViews) changes.RuleSetChanges {
```
xichen2020: `newTestRuleSetChanges`.

service/r2/store/kv/store_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	ruleSet, err := rules.NewRuleSetFromSchema(version, schema, rules.NewOptions())
+	if err != nil {
+		return nil, err
+	}
+
+	return ruleSet, nil
+}
+
+func emptyRuleSet(version int, meta rules.UpdateMetadata) (rules.RuleSet, error) {
```
xichen2020: `newEmptyTestRuleSet`.

generated/mocks/generate.go
```diff
@@ -0,0 +1,24 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// mockgen rules for generating mocks using file mode.
+//go:generate sh -c "mockgen -package=store -destination=$GOPATH/src/$PACKAGE/service/r2/store/storage_mock.go -source=$GOPATH/src/$PACKAGE/service/r2/store/store.go"
```
xichen2020: Perhaps just `$GOPATH/src/$PACKAGE/service/r2/store/mock.go` instead of `$GOPATH/src/$PACKAGE/service/r2/store/storage_mock.go` since it's under the `store` package so the name is self explanatory already?

service/r2/store/storage_mock.go
xichen2020: Yeah so the import ordering should be
```
stdlib imports

m3db imports

3rd party imports
```

service/r2/store/stub/store.go
```diff
@@ -21,14 +21,19 @@
 package stub
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
+	"github.com/pborman/uuid"
```
xichen2020: Import ordering.

service/r2/store/stub/store.go
```diff
@@ -595,4 +606,9 @@ func (s *store) FetchRollupRuleHistory(namespaceID, rollupRuleID string) ([]*mod
 	}
 }
 
+// This function is not supported. Use mocks package.
+func (s *store) UpdateRuleSet(rsChanges changes.RuleSetChanges, version int, uOpts r2store.UpdateOptions) (*models.RuleSet, error) {
```
xichen2020: Perhaps move this to be grouped together with other ruleset related methods?

service/r2/store/store.go
xichen2020: Yeah what I meant is that the convention is to simply call `s.ruleStore.ReadRuleSet(namespaceID)` in the store implementation. It doesn't seem like we need to call this outside the store package, so I'd prefer to remove it from the interface to keep the API surface area small.

service/r2/store/store.go
xichen2020: Yeah fair, can you please move this to be below `ValidateRuleSet` so the ruleset APIs are grouped together? (Also not sure why the `ValidateRuleSet` is in between namespace APIs, if you are up for adjusting that a little so namespace APIs are grouped together and ruleset APIs are grouped together, that'll be awesome).

nit: Please add a period at the end.

service/r2/store/kv/store.go
```diff
@@ -128,82 +131,86 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*mo
...
 	if err = s.ruleStore.WriteAll(nss, mutable); err != nil {
-		return s.handleUpstreamError(err)
+		return handleUpstreamError(err)
 	}
 
 	return nil
 }
 
-func (s *store) FetchRuleSet(namespaceID string) (*models.RuleSetSnapshotView, error) {
+func (s *store) FetchRuleSet(namespaceID string) (rules.RuleSet, error) {
```
xichen2020: Let's remove this API since we don't need to call it from outside the store package to keep API surface small.

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
 
 func (s *store) Close() { s.ruleStore.Close() }
 
-func (s *store) newUpdateMeta(uOpts r2.UpdateOptions) rules.UpdateMetadata {
+func (s *store) UpdateRuleSet(
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSet, error) {
+	rs, err := s.FetchRuleSet(rsChanges.Namespace)
```
xichen2020: Can do `rs, err := s.ruleStore.ReadRuleSet(namespaceID)` here like the others.

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
-		return nil, s.rollupRuleNotFoundError(namespaceID, rollupRuleID)
+		return nil, rollupRuleNotFoundError(namespaceID, rollupRuleID)
 	}
 	return hist, nil
 }
 
 func (s *store) Close() { s.ruleStore.Close() }
 
-func (s *store) newUpdateMeta(uOpts r2.UpdateOptions) rules.UpdateMetadata {
+func (s *store) UpdateRuleSet(
```
xichen2020: Can we move this to be below `ValidateRuleSet`?

service/r2/store/kv/store.go
xichen2020: Do we need to explicitly check this here? If there are changes after the user has sent up the proposed changes, the check and set will fail at the last step anyway no?

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
+	rs, err := s.FetchRuleSet(rsChanges.Namespace)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	if version != rs.Version() {
+		return nil, r2.NewConflictError("mismatched ruleset version number")
+	}
+
+	mutable := rs.ToMutableRuleSet().Clone()
+	err = mutable.ApplyRuleSetChanges(rsChanges, s.newUpdateMeta(uOpts))
```
xichen2020: Ah interesting I see, yeah I think you need to set the mutable ruleset version before apply ruleset changes, because that's the ruleset version used for check and set during a store operation in KV. You can add a `SetVersion` API for the `MutableRuleSet` interface and pull that change in here to do that.

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	err = s.ruleStore.WriteRuleSet(mutable)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+
+	snapshot, err := s.FetchRuleSetSnapshot(string(rs.Namespace()))
```
xichen2020: nit: you can use `rsChanges.Namespace` to avoid a `[]byte` to `string` conversion here.

Also conventionally we return the snapshot view (e.g., `FetchRuleSet` returns `*models.RuleSetSnapshotView`) so we should be consistent. You can always do the conversion to ruleset on the client side if needed.

public/r2/v1/swagger/swagger.json
```diff
@@ -184,42 +184,42 @@
...
-                        "in": "body",
-                        "name": "ruleset-version",
-                        "description": "The ruleset version to atomically apply bulk updates against",
-                        "required": true,
-                        "type": "integer"
+                        "in": "path",
+                        "name": "namespaceID",
+                        "description": "The name of the namespace",
+                        "type": "string",
+                        "required": true
```
xichen2020, line 222: Note this is a comment for the response, but I think for `200` response the schema for the result should be 
```
"responses": {
                    "200": {
                        "description": "The ruleset updates have been applied",
                        "schema": {
                            "$ref": "#/definitions/RuleSet"
                        }
                    },
```
Can you please fix?

service/r2/store/kv/store.go
xichen2020: Also thanks for moving them to the `store` package, makes sense.

service/r2/store/kv/store.go
jskelcy: This is not the case.  Because changes are applied to an in-memory dataset then committed to KV, at some point m3ctl will have to apply the changes before writing back to KV. So imagine the case where a user sends up some changes and while that request is inflight some new version of rules are landed. The users request will then make it to m3ctl and the updated ruleset will be pulled out so we can apply the changes onto the ruleset to write it back. If the user's changes are applied on top of the changes that just came out of KV there could be conflicts which will be blindly written over, and then sent to KV which will see the changes as valid because the version numbers match up.

Makefile
jskelcy: WTF. sorry my IDE corrects tabs or something.

service/r2/store/kv/store.go
xichen2020: Hm maybe I'm missing something here, in my mind the typical use case is:
```
1. users fetch latest ruleset, say version = v1 at this point
2. users make changes
3. users send changes alongside version v1
4. m3ctl commits changes by doing a checkandset against version v1.
```
If any thing changes in kv after step 1, that should move the version to v2 in KV which  should then cause a check and set failure when m3ctl tries to persist the changes no?

service/r2/store/storage_mock.go
jskelcy: RIght. But here I am not so worried about import ordering just more that this generated file does not seem to adhere to the standards we set, and if for generated code if we should just take what they give us, so every time you generate code it does not make changes to all the generated files.

service/r2/store/store.go
jskelcy: Good call. Does not need to be public right now.

service/r2/store/kv/store.go
jskelcy: well the KV api just allows for setting ruleset. So at some point we need to take ruleset and apply the changes to in application logic before we can commit back to KV. So we must fetch from KV to get a ruleset, apply the changes, and then write it back. But if the version in KV is different from the version the user was looking at when we created the alert, when we apply those changes in application logic we could blow away changes that happen after the user sent the request for changes. For an itemized view it looks more like this.

```
1. users fetch latest ruleset, say version = v1 at this point
2. users make changes
3. users send changes alongside version v1
4. m3ctl fetches changes and applies the changes to create a new ruleset (version 2).
5. m3ctl commits the new ruleset version 2 on top of version 1.
```
If in between stages 3 and 4 a different users updates what is in KV we dont want to try to apply the changes on top of the new version of the ruleset.

service/r2/store/kv/store.go
jskelcy: yeah let me do some digging. I dont think I was doing this in the past and it worked in staging, but I agree it does seem like I should need to update the version explicitly

service/r2/store/storage_mock.go
xichen2020: Oh I see did you pipe the generated code to mockclean? It's supposed to clean up the import order

generated/mocks/generate.go
xichen2020: Also I think you should be able to use reflection mode? If so, you can pipe the output to `mockclean` to fix the import ordering, e.g., `https://github.com/m3db/m3aggregator/blob/master/generated/mocks/generate.go#L22`.

service/r2/store/storage_mock.go
xichen2020: See comments in `generate.go`.

service/r2/store/kv/store.go
xichen2020: Yeah so I think the flow should be
```
1. users fetch latest ruleset, say version = v1 at this point
2. users make changes
3. users send changes alongside version v1
4. m3ctl fetches changes, set the version of the changes to v1, and apply the changes.
5. m3ctl commits the new ruleset iff the underlying version is still v1. If success, the underlying version will be automatically changed to v1+1 in KV.
```
Note the difference in how the version should be set in step 4. If in between stages 3 and 4 a different users updates what is in KV, the version in KV will be changed to `v1+1`, so in step when m3ctl tries to do a check and set against version `v1`, it will fail due to the underlying version having been changed.

service/r2/store/kv/store.go
jskelcy: It does not look like any of the other methods update the version by hand so I guess that is taken care of for me. Which means we would need to expose a new KV method that does not increment the version, which does not seem like a pattern we want to support.

xichen2020: I more meant we need to add a `SetVersion` method on the `MutableRuleSet` interface to set the version on the in-memory mutable ruleset (not the ruleset stored in kv) because that's the version used for check and set when persisting changes to kv, and then you can use it like below
```
rs, err := s.FetchRuleSet(rsChanges.Namespace)
if err != nil {
  return nil, handleUpstreamError(err)
}
mutable := rs.ToMutableRuleSet().Clone()
mutable.SetVersion(version)
if err = mutable.ApplyRuleSetChanges(rsChanges, s.newUpdateMeta(uOpts)); err != nil {
  return nil, handleUpstreamError(err)
}
if err = s.ruleStore.WriteRuleSet(mutable); err != nil {
  return nil, handleUpstreamError(err)
}
return s.FetchRuleSetSnapshot(rsChanges.Namespace)
```

jskelcy: We can sync up next week in person but I think that is problematic because you are fetching a potentially greater version than the one submitted by the user, setting the version to what the user set, applying the changes, so now you have a ruleset that should not really exist and then submitting it. It will fail which is what you want, but your intermediary ruleset is flawed and that seems kind of strange to me. 
But functionally both way of doing it get you to the same desired outcome.

xichen2020: Sure we can sync up, btw this is commonly how to make use of check and set based APIs, i.e., you fetch the change, apply the change **assuming** the fetched version is the desired version, and commit the change via check and set by checking the underlying version is the desired version. The end result is:
* If the underlying version is indeed the desired version, the change is successfully applied and the underlying version in KV is incremented.
* If the underlying KV version is changed after the user fetches the change and before the change is committed, the change will fail with a `StaleDataError`, and the user needs to retry. Since nothing is persisted, there is no harm done.

service/r2/store/kv/store.go
```diff
@@ -366,28 +373,59 @@ func (s *store) DeleteRollupRule(
...
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSet, error) {
+	rs, err := s.FetchRuleSet(rsChanges.Namespace)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	if version != rs.Version() {
+		return nil, r2.NewConflictError("mismatched ruleset version number")
```
xichen2020: Perhaps change the error to `r2.NewConflictError(fmt.Sprintf("ruleset version mismatch: current version=%d, expected version=%d", rs.Version(), version)` to be more explicit?

service/r2/store/kv/store.go
xichen2020: Also yeah I think both ways get to the desired outcome. I'm also fine if you want to keep it the way it is, if so please add some clarifying comments explaining why this would achieve the desired outcome, e.g., something alone the lines of
```
If the current KV version matches the desired version, the persistence operation will succeed if the underlying KV version remains the same before the change is committed and fail otherwise. If the current KV version doesn't match, we will fail fast.
```
since that's not immediately obvious.

jskelcy: I guess if I feel I have to leave a github comment on some code explaining how it works that is a good indicator that I should probably have a better in-code comment.

Makefile
```diff
@@ -18,6 +18,12 @@ metalint_exclude    := .excludemetalint
 m3ctl_package       := github.com/m3db/m3ctl
 gopath_prefix       := $(GOPATH)/src
 vendor_prefix       := vendor
+mocks_output_dir    := generated/mocks/mocks
+package_root        := github.com/m3db/m3ctl
+mocks_rules_dir     := generated/mocks
+auto_gen             := .ci/auto-gen.sh
```
xichen2020: These are still not aligned with the vars above, can you align all the vars?

generated/mocks/generate.go
xichen2020: Ping

service/r2/routes_test.go
xichen2020: Ping

service/r2/routes_test.go
xichen2020: Ping

service/r2/service.go
xichen2020: Ping

service/r2/service.go
xichen2020: Ping

service/r2/service.go
xichen2020: Ping

service/r2/store/kv/store.go
xichen2020: Ping

service/r2/store/kv/store_test.go
xichen2020: Ping

service/r2/store/kv/store_test.go
xichen2020: Ping

service/r2/store/kv/store_test.go
xichen2020: Ping

service/r2/service.go
```diff
@@ -94,6 +97,7 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
 		updateRollupRule:        instrument.NewMethodMetrics(scope, "updateRollupRule", samplingRate),
 		deleteRollupRule:        instrument.NewMethodMetrics(scope, "deleteRollupRule", samplingRate),
 		fetchRollupRuleHistory:  instrument.NewMethodMetrics(scope, "fetchRollupRuleHistory", samplingRate),
+		updateRuleSet:           instrument.NewMethodMetrics(scope, "udateRuleSet", samplingRate),
```
xichen2020: Typo: `udateRuleSet` -> `updateRuleSet`.

service/r2/store/kv/store.go
```diff
@@ -128,82 +130,115 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*mo
...
+func (s *store) UpdateRuleSet(
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSetSnapshotView, error) {
+	rs, err := s.fetchRuleSet(rsChanges.Namespace)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	// If ruleset version fetched from KV matches the change set version, the check and set operation will succeed if the underlying
```
xichen2020: nit: perhaps wrap the comments to make it shorter per line?

service/r2/store/kv/store.go
```diff
@@ -128,82 +130,115 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*mo
...
 
 	mr, exists := crs.MappingRules[mappingRuleID]
 	if !exists {
-		return nil, s.mappingRuleNotFoundError(namespaceID, mappingRuleID)
+		return nil, mappingRuleNotFoundError(namespaceID, mappingRuleID)
 	}
 	return mr, nil
 }
 
+func (s *store) UpdateRuleSet(
```
xichen2020: Can we move this to be below `ValidateRuleSet`? Grouping similar methods together makes it easier to navigate.

service/r2/store/kv/store.go
```diff
@@ -128,82 +130,115 @@ func (s *store) CreateNamespace(namespaceID string, uOpts r2.UpdateOptions) (*mo
...
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSetSnapshotView, error) {
+	rs, err := s.fetchRuleSet(rsChanges.Namespace)
```
xichen2020: Can we simply call `s.ruleStore.ReadRuleSet(rsChanges.Namespace)` here? Seems `fetchRuleSet` is only used here anyway so I'd be in favor of removing `fetchRuleSet`.

service/r2/store/kv/store.go
```diff
@@ -366,28 +401,32 @@ func (s *store) DeleteRollupRule(
...
-		return nil, s.rollupRuleNotFoundError(namespaceID, rollupRuleID)
+		return nil, rollupRuleNotFoundError(namespaceID, rollupRuleID)
 	}
 	return hist, nil
 }
 
 func (s *store) Close() { s.ruleStore.Close() }
 
-func (s *store) newUpdateMeta(uOpts r2.UpdateOptions) rules.UpdateMetadata {
+func (s *store) fetchRuleSet(namespaceID string) (rules.RuleSet, error) {
```
xichen2020: Let's inline this method where it's used (only in `UpdateRuleSet` it seems) and remove it to reduce cognitive burden.

service/r2/store/mock.go
```diff
@@ -0,0 +1,212 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
xichen2020: Hm strange that there is no license header here. How did you generate this? Did you use `make mock-gen`?

xichen2020 (commented):
Mostly LGTM, a few nits and comments

service/r2/store/mock.go
jskelcy: weird. Let me take a look at why that didnt work.

jskelcy: Looks like my local version of node is from 2014 or something. That seemed to cause issues, Ill make sure that is working.

jskelcy: It looks like the issue is because the file name changes. It used to be `storage_mock.go` but because we changed the file name to `mock.go`, and `auto-gen.sh` only looks for files with the `_mock` suffix the license is not added. Im going to change this file to be called `store_mock.go`

service/r2/store/kv/store.go
```diff
@@ -89,6 +89,39 @@ func (s *store) ValidateRuleSet(rs *models.RuleSetSnapshotView) error {
...
+	rsChanges changes.RuleSetChanges,
+	version int,
+	uOpts r2store.UpdateOptions,
+) (*models.RuleSetSnapshotView, error) {
+	rs, err := s.ruleStore.ReadRuleSet(rsChanges.Namespace)
+	if err != nil {
+		return nil, handleUpstreamError(err)
+	}
+	// If ruleset version fetched from KV matches the change set version,
+	//the check and set operation will succeed if the underlying
```
xichen2020: Space between `//` and `the`.

xichen2020 (approved):
LGTM

PR #72, opened 2018-08-15
Return 404 on not found errors when fetching a namespace which does not exist.

jskelcy:
Appears that a bit of the wiring was missing for returning 404s when a value is missing.


schallert:
LGTM

schallert (approved):

PR #74, opened 2018-08-20
Serve UI static assets from binary

robskillington:
This change includes the UI static assets directly in the binary itself and serves it from there so that the r2ctl HTTP service can be embedded into m3coordinator and m3dbnode itself.


prateek (approved):
LGTM

saminzadeh (approved):

PR #75, opened 2018-10-29
Update imports to monorepo

cw9:


prateek:
@cw9 maybe just migrate to monorepo too?

cw9:
Yeah, probably not now though, I want to fix this to unblock me from fixing the collectors. Also this repo feels like an app/extension built on top of `m3db/m3` with a bunch of UI, maybe this one can just be left outside.

prateek (approved):
SGTM

